PR #22, opened 2016-04-19
Inject rule

preslavsh:


mgechev:
This could be tricky to implement since some tokens can be injected **only** using `@Inject`, for instance:

```ts
const ANSWER = 42;

class AnswerLogger {
  constructor(@Inject(ANSWER) private _answer: number) {}
  log() {
    console.log(this._answer);
  }
}

const injector = Injector.createAndResolve([
  AnswerLogger,
  provide(ANSWER, { useValue: ANSWER })
]);

```

PR #25, opened 2016-04-20
refactor(rules): use more explicit naming

mgechev:
The rules now use the following naming:

```json
{
  "directive-selector-name": [true, "camelCase"],
  "component-selector-name": [true, "kebab-case"],
  "directive-selector-type": [true, "attribute"],
  "component-selector-type": [true, "element"],
  "directive-selector-prefix": [true, "sg"],
  "component-selector-prefix": [true, "sg"],
  "use-host-parameter-decorator": true,
  "use-input-parameter-decorator": true,
  "use-output-parameter-decorator": true,
  "no-attribute-parameter-decorator": true,
  "no-input-rename": true,
  "no-output-rename": true,
  "no-forward-ref" :true,
  "use-life-cycle-interface": true,
  "use-pipe-transform-interface": true,
  "pipe-naming": [true, "kebab-case","sg"]
}
```

The error messages are not yet finalized as well, since we're waiting for the styles' codes from the style guide.

We're planning to include a rule which verifies that a component selector has a valid custom element name.

Feedback for the rules' naming is welcome!

// cc @johnpapa @igorminar @preslavsh


preslavsh:
I like the new names of the rules.

johnpapa:
is `  "use-input-parameter-decorator": true,` referring to using the parameter to the `@Component`?

mgechev:
It refers to using the `@Input` parameter decorator:

```ts
@Component({
  selector: 'my-app',
  template: 'Hello {{name}}'
})
class AppComponent {
  @Input() name;
}
```

My arguments for this style are:

> The name of the property, or event name associated to `@Input` or respectively `@Output` should be modified only on a single place. In case of `inputs` decorator you need to modify both the inputs in `@Component`, and the declaration inside of the class's body.

> The metadata declaration attached to the directive is shorter and thus more readable.

However, there are also arguments against the style: definition of all the inputs in a single place is more intuitive/readable.

Probably this is not the most appropriate place for discussion of the style but I'm interested to hear your opinion. Also, what do you think about the name?

johnpapa:
I agree 100% to using the *property decorator* for these. I find it much more readable to see the decorator next to the property. Otherwise, when you scan a component class and see a property, we do not know which are input/outputs. 

I'm thinking about the name ...  perhaps go the other way ...  `use-input-output-property-decorator` ???

`use-input-parameter-decorator` feels confusions as it really is `use-component-decorators-input-parameter-decorator`

I'm also thinking it would be bad form to use output one way and input the other. So I'd tie these 2 together in a single setting

thoughts?

mgechev:
It should definitely be `property-decorator`. I'll rename it in a sec.

johnpapa:
should we combine those into 1 rule?


it would odd to set one to true and the other to false ... like this ....
```
"use-input-parameter-decorator": true,
"use-output-parameter-decorator": false,
```
That would be odd.

I'm thinking maybe ... 

```
"use-input-output-parameter-decorators": true,
```

PR #26, opened 2016-04-21
add no missed decorator rule and update README

preslavsh:


test/noMissedDecoratorRule.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+                    loader: ()=>System.import('./typescripts/app/company/load_applied_candidates').then(m=>m.AppliedCandidatesComponent),
+                    name: 'Candidates'
+                }),
+                new AsyncRoute({
+                    path: '/profile',
+                    loader: ()=>System.import('./typescripts/app/company/company_profile_form').then(m=>m.CompanyFormComponent),
+                    name: 'Profile'
+                })
+            ])
+            @View({
```
frederikschubert: `@View` is deprecated I think.

preslavsh:
Yes.
This rule has different problems and will be hard to implement.

PR #36, opened 2016-05-10
fix(util): kebab-case doesn't allow non-hyphen selector

### LINKED ISSUE #35: kebab-case shouldn't allow non-hyphened string
As [Style Guide](https://angular.io/styleguide#!#components) says, a component name should be named in kebab-case. And it should follow Custom Elements spec.

> Style 05-02
> Do use kebab-case for naming the element selectors of our components.
> Why? Keeps the element names consistent with the specification for [Custom Elements](https://www.w3.org/TR/custom-elements/).

Custom Elements spec says the following:

> The custom element type identifies a custom element interface and is a sequence of characters that must match the NCName production [XML-NAMES], must contain a U+002D HYPHEN-MINUS character, and must not contain any uppercase ASCII letters [HTML]. The custom element type must not be one of the following values:

But currently codelyzer allows a name doesn't contain hyphen as kebab-case.

- `selector: "foobar"` : passed
- `selector: "foo-bar"` : passed
- `selector: "fooBar"` : invalid

**Expected**

`selector: "foobar"` should be an error in kebab-case rule.

laco0416:
Close #35

### DIFF
--- a/src/util/selectorValidator.ts
+++ b/src/util/selectorValidator.ts
@@ -6,7 +6,7 @@ export const SelectorValidator = {
     return /^[^\[].+[^\]]$/.test(selector);
   },
   kebabCase(selector: string): boolean {
-    return /^[a-z0-9\-\[\]]+$/.test(selector);
+    return /^[a-z0-9\-]+\-[a-z0-9\-]+$/.test(selector);
   },
   camelCase(selector: string): boolean {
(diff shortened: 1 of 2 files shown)


laco0416:
Thanks!

mgechev:
Thanks for the PR! :-)

PR #50, opened 2016-06-03
Kebab case should validate on single lower-case word

ewancoder:
Kebab case is lowercase with words separated-by-dash. Therefore, single
lowercase word is still valid kebab-case.

Moreover, we have separate rules for validating mandatory dash:
- directive-selector-prefix
- component-selector-prefix
- pipe-naming

So for those of us who doesn't want to follow these it should be possible
to use single word selector name with kebab-case rule on.

### DIFF
--- a/src/util/selectorValidator.ts
+++ b/src/util/selectorValidator.ts
@@ -6,7 +6,7 @@ export const SelectorValidator = {
     return /^[^\[].+[^\]]$/.test(selector);
   },
   kebabCase(selector: string): boolean {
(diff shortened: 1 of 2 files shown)


mgechev:
The error message thrown by the rule is not correct.

On the other hand, we don't want to allow element selectors without any dash in their name since this is invalid custom element according to the spec.

In the perfect case we should have this PR merged and add one more rule which verifies if element selectors have dash in their names.

PR #59, opened 2016-06-14
feat(rules): add impure pipe rule

Kobzol:
Added a simple rule for checking impure pipes (the rule was derived from source of the pipe-naming rule).

### DIFF
--- a/src/pipeImpureRule.ts
+++ b/src/pipeImpureRule.ts
@@ -1,0 +1,61 @@
+import * as Lint from 'tslint/lib/lint';
+import * as ts from 'typescript';
+import {sprintf} from 'sprintf-js';
+import SyntaxKind = require('./util/syntaxKind');
+
+export class Rule extends Lint.Rules.AbstractRule {
+
+    public apply(sourceFile:ts.SourceFile):Lint.RuleFailure[] {
+        return this.applyWithWalker(
+            new ClassMetadataWalker(sourceFile, this));
+    }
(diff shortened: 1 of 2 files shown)


src/pipeImpureRule.ts
```diff
@@ -0,0 +1,61 @@
...
+import SyntaxKind = require('./util/syntaxKind');
+
+export class Rule extends Lint.Rules.AbstractRule {
+
+    public apply(sourceFile:ts.SourceFile):Lint.RuleFailure[] {
+        return this.applyWithWalker(
+            new ClassMetadataWalker(sourceFile, this));
+    }
+
+    static  FAILURE:string = 'The Pipe decorator of class %s should be pure.';
```
mgechev: Developer might be aware of what she is doing. I'd suggest to form the `FAILURE` message more like a warning rather than an error.

mgechev:
Looks awesome! I added a small comment for the error message.

src/pipeImpureRule.ts
Kobzol: That's right, this message is a bit too restrictive :-) I changed it, feel free to modify it to your liking, I'm not good with error messages :-)

Kobzol:
I modified the erorr message, and of course I forgot to fix the tests, so there are two new commits, you can squash them if you decide to merge this :-)

mgechev:
Looks good!

Issue #83, opened 2016-09-11
TypeError: Cannot read property 'kind' of undefined


rolandjitsu:
I get the following:
```bash
TypeError: Cannot read property 'kind' of undefined
    at /Users/rolandjitsu/Projects/infarm/dashboard/node_modules/codelyzer/selectorNameBase.js:79:50
    at Array.forEach (native)
    at SelectorNameValidatorWalker.validateSelector (/Users/rolandjitsu/Projects/infarm/dashboard/node_modules/codelyzer/selectorNameBase.js:77:18)
    at SelectorNameValidatorWalker.validateDecorator (/Users/rolandjitsu/Projects/infarm/dashboard/node_modules/codelyzer/selectorNameBase.js:70:18)
    at Array.forEach (native)
    at SelectorNameValidatorWalker.visitClassDeclaration (/Users/rolandjitsu/Projects/infarm/dashboard/node_modules/codelyzer/selectorNameBase.js:58:33)
    at SelectorNameValidatorWalker.SyntaxWalker.visitNode (/Users/rolandjitsu/Projects/infarm/dashboard/node_modules/tslint/lib/language/walker/syntaxWalker.js:264:22)
    at /Users/rolandjitsu/Projects/infarm/dashboard/node_modules/tslint/lib/language/walker/syntaxWalker.js:459:63
    at visitEachNode (/Users/rolandjitsu/Projects/infarm/dashboard/node_modules/typescript/lib/typescript.js:8247:30)
    at Object.forEachChild (/Users/rolandjitsu/Projects/infarm/dashboard/node_modules/typescript/lib/typescript.js:8405:24)
    at SelectorNameValidatorWalker.SyntaxWalker.walkChildren (/Users/rolandjitsu/Projects/infarm/dashboard/node_modules/tslint/lib/language/walker/syntaxWalker.js:459:12)
    at SelectorNameValidatorWalker.SyntaxWalker.visitBlock (/Users/rolandjitsu/Projects/infarm/dashboard/node_modules/tslint/lib/language/walker/syntaxWalker.js:28:14)
    at SelectorNameValidatorWalker.SyntaxWalker.visitNode (/Users/rolandjitsu/Projects/infarm/dashboard/node_modules/tslint/lib/language/walker/syntaxWalker.js:249:22)
    at /Users/rolandjitsu/Projects/infarm/dashboard/node_modules/tslint/lib/language/walker/syntaxWalker.js:459:63
[...]

PR #85, opened 2016-09-14
update validation of arg

preslavsh:
Smart guess what is the problem in #83 . My point is that based on the stack trace, the error occurs in 
the class selectorNameBase, method validateSelector. Usually, this kind of errors, are created, when there is a problem with the parsing of a given ts file into AST. I can not be sure, that is right or wrong because the issue creator has signed NDA and cannot provide us code samples

### DIFF
--- a/src/selectorNameBase.ts
+++ b/src/selectorNameBase.ts
@@ -69,7 +69,7 @@ class SelectorNameValidatorWalker extends Lint.RuleWalker {
   }
 
   private validateSelector(className: string, arg: ts.Node) {
-    if (arg.kind === SyntaxKind.current().ObjectLiteralExpression) {
+    if (arg && arg.kind === SyntaxKind.current().ObjectLiteralExpression) {
       (<ts.ObjectLiteralExpression>arg).properties.filter(prop => (<any>prop.name).text === 'selector')
       .forEach(prop => {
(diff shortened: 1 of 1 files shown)


src/selectorNameBase.ts
```diff
@@ -69,7 +69,7 @@ class SelectorNameValidatorWalker extends Lint.RuleWalker {
   }
 
   private validateSelector(className: string, arg: ts.Node) {
-    if (arg.kind === SyntaxKind.current().ObjectLiteralExpression) {
+    if (arg && arg.kind === SyntaxKind.current().ObjectLiteralExpression) {
```
rolandjitsu: @preslavsh would this also not mean that the selectors are not going to be validated anymore?

preslavsh: Yes, but I cannot react fast, without seeing code snippets and you can not provide me some, which gives me only this option for now.

 I will invest more time, solving the issue, during the weekend.

rolandjitsu: Say I could provide some code, I have 200+ files, how would I know which one is causing this? I cannot just give you access to everything. Maybe you can point me to some things I could look at in the components that could cause this? It might be that is not even related to the actual components but the linter or TS.

PR #102, opened 2016-09-28
Missing properties fix

troyanskiy:
It's related to that issue https://github.com/mgechev/angular2-seed/issues/1400
I'm not sure that I'm fixing correctly that 'bug'

### DIFF
--- a/src/angular/ng2Walker.ts
+++ b/src/angular/ng2Walker.ts
@@ -138,6 +138,9 @@ export class Ng2Walker extends Lint.RuleWalker {
     let name = getDecoratorName(decorator);
     if (name === 'Component') {
       this.visitNg2Component(<ts.ClassDeclaration>decorator.parent, decorator);
+      if (!(<ts.ObjectLiteralExpression>(<ts.CallExpression>decorator.expression).arguments[0]).properties) {
+        return;
+      }
       const inlineTemplate = (<ts.ObjectLiteralExpression>
           (<ts.CallExpression>decorator.expression).arguments[0])
             .properties.map((prop: any) => {


src/angular/ng2Walker.ts
```diff
@@ -138,6 +138,9 @@ export class Ng2Walker extends Lint.RuleWalker {
     let name = getDecoratorName(decorator);
     if (name === 'Component') {
       this.visitNg2Component(<ts.ClassDeclaration>decorator.parent, decorator);
+      if (!(<ts.ObjectLiteralExpression>(<ts.CallExpression>decorator.expression).arguments[0]).properties) {
```
mgechev: With the following change I get the same error for this test case:

```ts
  it('should ignore empty @Component decorator', () => {
    let source = `
      @Component()
      class Baz {}
    `;
    let ruleArgs: tslint.IOptions = {
      ruleName: 'foo',
      ruleArguments: ['foo'],
      disabledIntervals: null
    };
    let sf = ts.createSourceFile('foo', source, null);
    let walker = new Ng2Walker(sf, ruleArgs);
    (<any>chai).expect(() => {
      let templateSpy = chaiSpy.on(RecursiveAngularExpressionVisitor.prototype, 'visitPropertyRead');
      walker.walk(sf);
      (<any>chai.expect(templateSpy).to.not.have.been).called();
    }).not.to.throw();
  });
```

If I change the check to:

```ts
// ...
if (!(<ts.CallExpression>decorator.expression).arguments.length || 
    !(<ts.ObjectLiteralExpression>(<ts.CallExpression>decorator.expression).arguments[0]).properties) {
  return;
}
// ...
```

It works fine

mgechev:
@troyanskiy thanks for the help again! I closed the issue and added this fixed, which is based on your PR https://github.com/mgechev/codelyzer/pull/105.

troyanskiy:
@mgechev I'm so sorry that I did not add tests for the case. I did not have time for that. Thanks you for understanding.

PR #113, opened 2016-10-11
fix(pipeNaming): do not fail on empty @Pipe decorator

### LINKED ISSUE #111: Error on `@Pipe` decorator
```ts
@Pipe
export class Foo implements PipeTransform {}
```

```
TypeError: Cannot read property 'text' of undefined
    at node_modules/codelyzer/pipeNamingRule.js:52:39
```

mgechev:
Add `visitNg2Pipe` method in `NgwWalker`.
Refactor pipe related rules to use `Ng2Walker`.

Fix #111

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,7 @@ os:
 branches:
   only: master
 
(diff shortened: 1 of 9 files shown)


mgechev:
Incorporated in #116

Issue #133, opened 2016-11-03
New linting rules to fail if a test is focused or skipped


Maloric:
It is fairly common in our team to focus on an individual test or describe in order to run it quickly whilst working on that part of the system.  However we don't want to commit this focused test as it may cause the build to pass whilst only running a subset of the tests.  This lint rule ensures that such a build would fail, because we do not allow focused tests or skipped tests to be pushed to the repository.

Such a rule would surely be useful to others, so I have forked Codelyzer and added the two rules (plus unit tests), and would like to create a pull request for this feature if anyone thinks it would be useful.

Maloric:
I realised that this repo is not the right place for this issue.

PR #134, opened 2016-11-03
Added no-focused-test and no-skipped-test rules

Maloric:
This pull request adds two new linting rules to disallow focused tests and skipped tests (fit, fdescribe, xit and xdescribe), as per issue #133.


Maloric:
I realised that this repo is not the right place for this pull request.

PR #137, opened 2016-11-05
Selectors

preslavsh:


src/componentSelectorRule.ts
```diff
@@ -0,0 +1,143 @@
...
+    constructor(ruleName: string, value: any, disabledIntervals: Lint.IDisabledInterval[]) {
+        let prefix = value[2];
+        let name = value[3];
+        super(ruleName, value, disabledIntervals);
+        this.setMultiPrefix(prefix);
+        this.setPrefixArguments(prefix);
+        this.setPrefixValidator(prefix);
+        this.setPrefixFailure();
+        if (name === 'kebab-case') {
+            this.nameValidator = SelectorValidator.kebabCase;
```
mgechev: If it is `camelCase` the value of the validator is not set.

src/directiveSelectorRule.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    constructor(ruleName: string, value: any, disabledIntervals: Lint.IDisabledInterval[]) {
+        let prefix = value[2];
+        let name = value[3];
+        super(ruleName, value, disabledIntervals);
+        this.setMultiPrefix(prefix);
+        this.setPrefixArguments(prefix);
+        this.setPrefixValidator(prefix);
+        this.setPrefixFailure();
+        if (name === 'camelCase') {
```
mgechev: If the value is `kebab-case` the validator is not set.

src/directiveSelectorRule.ts
```diff
@@ -0,0 +1,143 @@
...
+        let expr = baseExpr.expression || {};
+        let name = expr.text;
+        let args = baseExpr.arguments || [];
+        let arg = args[0];
+        if (name === 'Directive') {
+            this.validateSelector(className, arg);
+        }
+    }
+
+    private validateSelector(className: string, arg: ts.Node) {
```
mgechev: Same comment as above. This can be simplified by filtering only properties with supported types.

mgechev:
LGTM!

PR #150, opened 2016-11-15
chore: migrate to angular 2.2.0

ocombe:
I cannot update my project because codelyzer peer dependencies are not compatible with the current version, let's hope this gets merged and released soon.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,8 +43,8 @@
   },
   "homepage": "https://github.com/mgechev/codelyzer#readme",
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


ocombe:
Ah damn it looks like it's not as simple as updating the dependencies, I'll close this PR and let you do a proper update of the lib then

mgechev:
Thanks for pointing this out! Here's the issue and the fix https://github.com/mgechev/codelyzer/issues/152.

PR #177, opened 2016-12-05
Atom fatal error from app-root-path

scttcper:
In Atom Editor when using linter-tslint and codelyzer 2.0.0-beta.1 we're unable to lint files currently because app-root-path fails. I'm not sure why? But putting it inside the try catch to fail silently stops it from being broken.


mgechev:
Thanks for the PR! I'll include it as part of beta.2, which should be released in the next week or two.

PR #180, opened 2016-12-07
[Experimental] First attempt at replacing AST traversal with Maybe<T>

ghsyeung:
**Prototyping of Maybe monad in AST traversal**

This is intended to further discuss whether/how Maybe monad would be a good fit in 
  * writing internal AST traversal
  * jQuery style helpers for Rules development

src/angular/metadataReader.ts seems to be a good target to start. Most of the work goes into extracting AST traversal code into `astQuery` and `ngQuery`.

A few realizations so far
  * Maybe monad introduces a bit a learning curve, but the structure makes the intent easier to understand (e.g. `decoratorArgument`)
  * There are a few wrinkles getting the proper type returned, but type guard helps (e.g. `callExpression`)


mgechev:
Looks hot, happy to see your suggestion in action! Will take a look at it in the next a couple of days and give you feedback. Thanks!

ghsyeung:
Had some more thoughts about refactor experience so far.

Can codelyzer develop against a different TS than the runtime TS? I have a feeling you can, but if not, it harder to write compatible type-friendly code as a lot of useful features like 'control flow based type analysis' and 'dot name type guards' landed only in TS 2.0

More refactoring can be done in `src/angular/metadataReader.ts`, but that likely require adding a 'seam' to urlResolver and fileResolver to make it Maybe friendly.

Also there may be a few minor bugs, I'll highlight them below.

src/util/astQuery.ts
```diff
@@ -0,0 +1,75 @@
...
+    return Maybe.lift(dec.expression)
+        .fmap(expr => isCallExpression(expr) ? expr as ts.CallExpression : undefined);
+}
+
+export function isPropertyAssignment(expr: ts.ObjectLiteralElement): expr is ts.PropertyAssignment {
+    return expr && expr.kind === kinds.PropertyAssignment;
+}
+
+export function isSimpleTemplateString(expr: ts.Expression): expr is (ts.StringLiteral | ts.NoSubstitutionTemplateLiteral) {
+    return expr && expr.kind === kinds.StringLiteral || expr.kind === kinds.NoSubstitutionTemplateLiteral;
```
ghsyeung: I believe https://github.com/mgechev/codelyzer/blob/master/src/util/utils.ts#L29 should use `kinds.NoSubstitutionTemplateLiteral` instead.

I was using astExplorer and find it odd that it maps to `FirstTemplateToken` as well.

test/angular/metadataReader.spec.ts
```diff
@@ -171,12 +171,14 @@ describe('metadataReader', () => {
     it('should work invoke Config.resolveUrl after all resolves', () => {
       let invoked = false;
       const bak = Config.resolveUrl;
-      Config.resolveUrl = (url: string) => {
-        invoked = true;
-        chai.expect(url.startsWith(normalize(join(__dirname, '../..')))).eq(true);
-        return url;
-      };
-      const code = `
+      try {
```
ghsyeung: I think we need to wrap with try/finally or the spy would leak when a test fails.

ghsyeung:
Added Walker builder (a builder approach and a functional approach) in attempt to make Rule definition easier.

src/angular/metadataReader.ts
```diff
@@ -1,162 +1,155 @@
...
-      return this.readDirectiveMetadata(d, directiveDecorator);
+    constructor(private _fileResolver: FileResolver, private _urlResolver?: AbstractResolver) {
+        this._urlResolver = this._urlResolver || new UrlResolver(new PathResolver());
     }
-    if (componentDecorator) {
-      return this.readComponentMetadata(d, componentDecorator);
+
+    read(d: ts.ClassDeclaration): DirectiveMetadata {
+        let componentMetadata = unwrapFirst(
+            (d.decorators || ([] as ts.Decorator[])).map((dec: ts.Decorator) => {
```
mgechev: Looks beautiful!

src/angular/metadataReader.ts
```diff
@@ -1,162 +1,155 @@
...
-            url: null,
-            node: inlineStyle,
-          });
+
+    readComponentStylesMetadata(dec: ts.Decorator, external: MetadataUrls) {
+        let styles: any[];
+        const inlineStyles_M = getInlineStyle(dec)
+            .fmap(inlineStyles => {
+                return inlineStyles.elements.map((inlineStyle: ts.Expression) => {
+                    if (isSimpleTemplateString(inlineStyle)) {
```
mgechev: Maybe we can first apply `filter` and after that map it, so we can drop the ugly if condition?

src/angular/metadataReader.ts
```diff
@@ -1,162 +1,155 @@
...
     }
-    return styles;
-  }
-
-  readComponentMetadata(d: ts.ClassDeclaration, dec: ts.Decorator) {
-    const expr = this.getDecoratorArgument(dec);
-    const metadata = this.readDirectiveMetadata(d, dec);
-    const result = new ComponentMetadata();
-    result.selector = metadata.selector;
-    result.controller = metadata.controller;
```
mgechev: Seems like an issue.

src/util/function.ts
```diff
@@ -0,0 +1,81 @@
...
+export function unwrapFirst<T>(ts: Maybe<T>[]): T|undefined {
+    const f = ts.find((t: Maybe<T>) => t.isSomething);
+    if (!!f) {
+        return f.unwrap();
+    }
+    return undefined;
+}
+
+export function all<T>(...preds: F1<T, boolean>[]): F1<T, boolean> {
+    return (t: T) => !preds.find(p => !p(t));
```
mgechev: Maybe `some` can do the trick since it returns boolean anyway.

src/util/function.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+export function any<T>(...preds: F1<T, boolean>[]): F1<T, boolean> {
+    return (t: T) => !!preds.find(p => p(t));
```
mgechev: Maybe `some` can do the trick since it returns boolean anyway.

mgechev:
I left some comments. The linter is failing:

```
src/componentClassSuffixRule.ts(15,31): error TS2304: Cannot find name 'suffix'.
src/componentClassSuffixRule.ts(25,32): error TS2334: 'this' cannot be referenced in a static property initializer.
src/componentClassSuffixRule.ts(25,37): error TS2339: Property 'getOptions' does not exist on type 'typeof Rule'.
src/componentClassSuffixRule.ts(25,63): error TS2334: 'this' cannot be referenced in a static property initializer.
src/componentClassSuffixRule.ts(25,68): error TS2339: Property 'getOptions' does not exist on type 'typeof Rule'.
src/componentClassSuffixRule.ts(26,61): error TS2346: Supplied parameters do not match any signature of call target.
```

ghsyeung:
Nothing new yet, just fixed the above error resulted from merging master.

mgechev:
Awesome! I just published beta.4, we can introduce this new AST traversal mechanism in beta.5.

PR #197, opened 2016-12-16
chore(dependencies) upgrade tslint to 4.1

nexus-uw:
tslint 4.1 was released yesterday https://github.com/palantir/tslint/blob/master/CHANGELOG.md#v410


mgechev:
Maybe a better idea is to use tslint@^4.0.0. Would you update the PR?

nexus-uw:
updated.

mgechev:
Thanks for the PR!

PR #202, opened 2016-12-20
chore: support angular 2.4

clydin:


package.json
```diff
@@ -59,16 +59,16 @@
...
     "ts-node": "1.2.2",
     "tslint": "^4.0.0",
     "typescript": "^2.1.4",
     "zone.js": "^0.7.2"
   },
   "peerDependencies": {
     "tslint": "^4.0.0",
-    "@angular/compiler": "<=2.3 >=2.2.0",
-    "@angular/core": "<=2.3 >=2.2.0"
+    "@angular/compiler": "<=2.4 >=2.2.0",
```
mhoyer: I am wondering if there is concrete intent to have this strict dependency regarding upper version boundary? Cause this will always produce a time frame where codelyzer is hanging after a new ng2 release. 

Wouldn't ```"<3.0 >=2.2.0"``` be better?

mgechev: I'm using some internal APIs of the compiler which often change between minor versions.

mhoyer: I see. So it is. ;-)

Issue #221, opened 2017-01-24
Compilation with `declaration` enabled doesn't work


mgechev:
After the recent refactoring the generation of type definitions stopped working.

@ghsyeung would you have chance to take a look at it?

ghsyeung:
will take a look tonight

ghsyeung:
how can I reproduce? "npm run release"?

mgechev:
Thanks! Yes, I'm assigning it to you.

ghsyeung:
Took a look last night, still deciding how not to leak implementation types
at the library level.
I should have something ready soon.

PR #223, opened 2017-01-27
Updating visibility and export to fix d.ts build error

ghsyeung:
Fixes errors from building with declaration file (#221)

Tested with `npm run tsc -- -p tsconfig-release.json`.

Making `walkerBuilder` private as it should be local to the Rule class.


mgechev:
Thanks for the PR! I'll review & merge as first thing tomorrow!

PR #239, opened 2017-02-19
fix(templatesUsePublicRule): Pass member name back through error

### LINKED ISSUE #229: Better logging of errors required
With the new package we get errors like 'You can bind only to public class members.' with occur to happen at line 1, column 1 of the HTML template. But the template has `<div class="main-menu-button"> `on the first line. 

If I look to the code of the templatesUsePublicRule the specific member is known. So could we at least have the name of the member or method that is causing the error in the error message and not a generic error?

comfroels:
Pass the class member name to the error for more clarification

fixes #229

### DIFF
--- a/src/templatesUsePublicRule.ts
+++ b/src/templatesUsePublicRule.ts
@@ -41,7 +41,7 @@ class SymbolAccessValidator extends RecursiveAngularExpressionVisitor {
         .some(m => m.kind === SyntaxKind.current().PrivateKeyword || m.kind === SyntaxKind.current().ProtectedKeyword);
       const width = ast.name.length;
       if (!isPublic) {
-        const failureString = 'You can bind only to public class members.';
+        const failureString = `You can bind only to public class members. ${member.name.getText()} is not a public class member.`;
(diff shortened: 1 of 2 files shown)


src/templatesUsePublicRule.ts
```diff
@@ -41,7 +41,7 @@ class SymbolAccessValidator extends RecursiveAngularExpressionVisitor {
         .some(m => m.kind === SyntaxKind.current().PrivateKeyword || m.kind === SyntaxKind.current().ProtectedKeyword);
       const width = ast.name.length;
       if (!isPublic) {
-        const failureString = 'You can bind only to public class members.';
+        const failureString = `You can bind only to public class members. ${member.name.getText()} is not a public class member.`;
```
mgechev: Would you surround it in quotes in order to be consistent with other rules?

mgechev:
@comfroels awesome job! Thanks! I left only a single comment.

comfroels:
Yeah not a problem! One of my first OSS contributions.

mgechev:
Thanks again! I'll update the list of contributors in the next few days.

Issue #242, opened 2017-02-20
Drop unused CSS selectors


mgechev:
Currently we mark CSS as unused only when all the rules for the block are unused.

In a better 🌍 we'd also mark individual selectors which are unused.

PR #248, opened 2017-02-22
No unused css

comfroels:
#242 
Not sure if this is exactly what is needed or not. Not very comfortable with the codebase yet. Let me know what else needs to get added or removed for this to work. Also, I would like to know how to hook up a local project to my local copy of codelyzer? I think it will help me test a little better.


mgechev:
Thanks for the PR! I'll review it during the weekend and give you feedback.

What I meant in the issue was to report unused selectors like here:

```
template: `
  <div class="foo">
  </div>
  <div class="bar">
  </div>
`,
styles: [
  `
    .foo,
    .baz {
      color: red;
    }
  `
]
```

`baz` need to be underlined as unused, also the autofix should work and remove: `\n    .baz`.

mgechev:
@comfroels it seems this implementation does something else. Do you want to close this PR or you'd want to work here?

comfroels:
I'll close it. I'll take another shot at it later this week. I think I've read the source code a little more and understand it a little more

mgechev:
@comfroels thanks for the effort! This functionality will require changes only in `noUnusedCss.ts` and of course `noUnusedCss.spec.ts`.

comfroels:
Sounds good.

PR #254, opened 2017-03-01
Allow tslint to run with codelizer without node_modules in the root

rtfpessoa:
Like most of the other TSLint plugins it would be great to be able to run tslint with codelizer without the required node_modules directory. Ideally using codelizer from the path.

This PR exports the rules, allowing to use the `extends` field in tslint.json. This is also similar to the way ESLint plugins work.

As an example you can check https://github.com/buzinas/tslint-eslint-rules/blob/master/index.js

This PR still allows for the previous way without breaking compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,51 @@ You can easily use codelyzer with your custom setup:
 npm i codelyzer@2.0.0-beta.4 tslint@4.0.0 typescript@2.0.9 @angular/core@2.4.0 @angular/compiler@2.4.0 rxjs@5.0.1 zone.js@0.7.2
 ```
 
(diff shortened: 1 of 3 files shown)


rtfpessoa:
@mgechev I think I might have forgotten something. From what I saw seems like some files are removed. Is `index.js` going to be released to npm?

mgechev:
Yes, `index.js` is pointed by `main`. What did you forget?

rtfpessoa:
I was not sure if you were excluding some files from the project at publish time. In the local codelyzer folder for older versions I only saw JS files in the root I was not sure if I needed to add this new file somewhere for it to be included at publish time.

PR #259, opened 2017-03-06
docs(testHelper): explain test helper functions

lazarljubenovic:
So it took me a while to understand how tests work (and tests are the best place to start looking at when digging at source code for the first time), so I figured I'd help others understand what functions are used for.

If I misunderstood something please comment so I can edit the explanation.

(Or, if we disagree about having these kinds of "docs", maybe we could open a wiki "guide/tutorial" about understanding the code and going towards a first PR?)

---

Also added an `.editorconfig` file with just two basic rules for now (I'll go though other files in a different PR tonight).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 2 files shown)


mgechev:
Thanks for the docs!

PR #261, opened 2017-03-07
feat(rules): disallow inapplicable lifecycles

### LINKED ISSUE #126: feature: if class uses lifecycle hooks it must had proper name decorator
For example, `ngOnInit` or `ngOnDestroy` can be used on service, but it is component/directive hooks.

``` typescript
import {
    Injectable,
    OnInit,
    OnDestroy,
} from '@angular/core';

@Injectable()
export class MyService implements OnInit, OnDestroy {
    ngOnInit() {
    }

    ngOnDestroy() {
    }
}
```

Lifecycle sequence
https://angular.io/docs/ts/latest/guide/lifecycle-hooks.html
Some are "component-only", some - "directive/component".

Variant 1
- show error when class have not `@Component` or `@Directive` decorator

Variant 2
- show error when class have `@Injectable` decorator, because it is not work on services

lazarljubenovic:
closes #126

### DIFF
--- a/src/noInapplicableLifeCycleRule.ts
+++ b/src/noInapplicableLifeCycleRule.ts
@@ -1,0 +1,159 @@
+import * as Lint from 'tslint';
+import * as ts from 'typescript';
+import * as SyntaxKind from './util/syntaxKind';
+import {sprintf} from 'sprintf-js';
+import {Ng2Walker} from './angular/ng2Walker';
+import {difference, intersection} from './util/setOperations';
+import {InheritanceForest} from './util/inheritanceForest';
+
+
+function getHeritageNames(t: any): string {
+  if (t.expression && t.expression.name) {
+    return t.expression.name.text;
+  }
+  return t.expression.text;
+}
+
+// TODO
+// I get
+// error TS2365: Operator '==' cannot be applied to types 'SyntaxKind' and 'SyntaxKind'.
+// if I specify syntax type as SyntaxKind.SyntaxKind
+function getHeritage(node: ts.ClassDeclaration, syntax: any): any[] {
+  if (node.heritageClauses) {
+    const clause = node.heritageClauses
+      .filter((clause: ts.HeritageClause) => clause.token == syntax);
+    if (clause.length) {
+      return clause[0].types;
+    }
+  }
+  return [];
+}
+
+function getDecoratorNames(node: ts.ClassDeclaration): string[] {
+  if (node.decorators) {
+    return node.decorators.map((d: any) => {
+      if (d.expression.expression) {
+        // If decorator is invoked; @Injectable()
+        return d.expression.expression.text;
+      } else {
+        // If decorator is not invoked; @Injectable
+        return d.expression.text;
+      }
+    });
+  }
+  return [];
+}
+
+export class Rule extends Lint.Rules.AbstractRule {
+
+  static FAILURE_NOT_CMP_OR_DIR_FACTORY = (hookNames: {name: string, path: string[] | null}[]) => {
+    const commonMessage = `The %s class is not a Component or Directive yet implements`;
+    const hooksWithPaths = hookNames
+      .map(hook => {
+        let str = hook.name;
+        if (hook.path) {
+          str += ` (from ${hook.path.slice(1, -1).join(' -> ')})`;
+        }
+        return str;
+      });
+    if (hooksWithPaths.length == 1) {
+      return `${commonMessage} a life cycle hook ${hooksWithPaths[0]}.`;
+    } else {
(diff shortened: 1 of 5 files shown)


src/noInapplicableLifeCycleRule.ts
```diff
@@ -0,0 +1,109 @@
...
+
+
+function getInterfaceNames(t: any): string {
+  if (t.expression && t.expression.name) {
+    return t.expression.name.text;
+  }
+  return t.expression.text;
+}
+
+function getInterfaces(node: ts.ClassDeclaration): any[] {
```
mgechev: The behavior could be incorrect in case this class is base for a component, i.e.:

```
class A implements OnInit {...}

@Component(...)
class B extends A {...}
```

mgechev:
Thanks for the PR! There are some corner cases which we should discuss. The first one (implementing an interface in a base class), is not something which can be solved trivially.

src/noInapplicableLifeCycleRule.ts
lazarljubenovic: So if your example was just he first line, it would fail because `A` is not a component.

But _becase_ it's extended, it's fine to have a class which implements `OnInit`. So as long as it's used somewhere _down_ the inheritence chain, implementing is not a lint error.

Am I following?

mgechev: Yes, correct.

src/noInapplicableLifeCycleRule.ts
```diff
@@ -0,0 +1,109 @@
...
+    if (clause.length) {
+      return clause[0].types;
+    }
+  }
+  return [];
+}
+
+function getDecoratorNames(node: ts.ClassDeclaration): string[] {
+  if (node.decorators) {
+    return node.decorators.map((d: any) => d.expression.expression.text);
```
lazarljubenovic: Yup, fixed this one.

src/noInapplicableLifeCycleRule.ts
```diff
@@ -0,0 +1,159 @@
...
+    return t.expression.name.text;
+  }
+  return t.expression.text;
+}
+
+// TODO
+// I get
+// error TS2365: Operator '==' cannot be applied to types 'SyntaxKind' and 'SyntaxKind'.
+// if I specify syntax type as SyntaxKind.SyntaxKind
+function getHeritage(node: ts.ClassDeclaration, syntax: any): any[] {
```
lazarljubenovic: Unsure why this happens.

src/noInapplicableLifeCycleRule.ts
```diff
@@ -0,0 +1,159 @@
...
+    // Remove an earlier added possible failure
+    heritageNames.forEach(base => {
+      if (this.possibleFailures.has(base)) {
+        // We marked this as a possible failure
+        // Now we know it's not because it's extended (so the error is "propagated"  down the
+        // inheritance tree until the leaves where it's either resolved or stays there
+        this.possibleFailures.delete(base);
+      }
+    });
+    this['failures'] = []; // Clear failures?
```
lazarljubenovic: So I couln't find any hook or a way to collect some data and after walking is complete to do something with it. This method is called by `applyWalker` and it just returns the failures. I can see how to get the failures, but not how to change them after I've added them.

test/noInapplicableLifeCycleRule.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+                               ~
+      `;
+      const message = `The A class is not a Component or Directive yet implements these life cycle hooks: ` +
+        `OnInit (from Base -> VeryBase), OnDestroy (from Base -> VeryBase), AfterViewInit (from Base).`;
+      assertAnnotated({ruleName, source, message});
+    });
+  });
+
+  describe(`multiple inheritance paths`, () => {
+    // TODO: merge these two in a single test (blocked by assert annotated working with multiple failures)
```
lazarljubenovic: I can try this in a separate PR.

Something like: 

```
assertMultipleAnnotated({ruleName, source,
  [ { char: '~', msg: 'Failure 1' },
    { char: '#', msg: 'Failure 2' } ])
```

mgechev: Another alternative is what tslint does [here](https://github.com/palantir/tslint/blob/master/test/rules/jsdoc-format/jsdoc.js.lint).

src/noInapplicableLifeCycleRule.ts
lazarljubenovic: @mgechev Can you point me in the right direction? Do we have a rule which does something like this that I can look at? I have a feeling it's super obvious but I'm totally lost :sweat_smile:

PR #266, opened 2017-03-08
docs(meta): add metadata to all rules

lazarljubenovic:
I realized too late that some of the rules are not in v3 anymore, but oh well. I tried quoting the official styleguide whenever it made sense in the context.

I've seen `Lint.Utils.dedent` in tslint source code so no idea how exactly it works, but we'll see after building the docs page and change it accordingly later, I guess.

### DIFF
--- a/src/invokeInjectableRule.ts
+++ b/src/invokeInjectableRule.ts
@@ -4,6 +4,17 @@ import {sprintf} from 'sprintf-js';
 import {Ng2Walker} from './angular/ng2Walker';
 
 export class Rule extends Lint.Rules.AbstractRule {
+  public static metadata: Lint.IRuleMetadata = {
+    ruleName: 'invoke-injectable',
+    type: 'functionality',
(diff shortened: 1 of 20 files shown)


src/invokeInjectableRule.ts
```diff
@@ -4,6 +4,17 @@ import {sprintf} from 'sprintf-js';
 import {Ng2Walker} from './angular/ng2Walker';
 
 export class Rule extends Lint.Rules.AbstractRule {
+  public static metadata: Lint.IRuleMetadata = {
+    ruleName: 'directive-class-suffix',
```
mgechev: It's `invoke-injectable`.

mgechev:
Great job! Thanks for the PR! I left only one comment for wrong rule name.

src/invokeInjectableRule.ts
lazarljubenovic: Fixed.

PR #267, opened 2017-03-11
test(*): support multiple annotated failures

lazarljubenovic:
The [way `tslint` does it](https://github.com/palantir/tslint/blob/master/test/rules/jsdoc-format/jsdoc.js.lint) seems like an overkill and is actually less readable for our small tests (too much clutter). It looks much clearer to me like this. We can add the `[failure]` syntax when we feel the need for it. There are enough symbols like this; it basically it just has to be a symbol which is not otherwise used in the snippet.

I updated a (kinda) random test to test the new helper function, although it doesn't make _that_ much sense to test two same error messages (I couldn't find an existing rule on master branch in which two different error messages can happen in the same context).

### DIFF
--- a/test/componentSelectorRule.spec.ts
+++ b/test/componentSelectorRule.spec.ts
@@ -1,4 +1,4 @@
-import { assertSuccess, assertAnnotated} from './testHelper';
+import {assertSuccess, assertAnnotated, assertMultipleAnnotated} from './testHelper';
 
 describe('component-selector-prefix', () => {
     describe('invalid component selectors', () => {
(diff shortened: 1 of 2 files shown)


mgechev:
What do you think about suffixing the markers with an identifier? Something like:

```
@Component({ selector: '[bar]' })
                                          ~~~~[1]
@Component({ selector: '[foo]' })
                                          ~~~~[2]
```

This way, we can verify the failure like:

```
failures: {
  1: '...',
  2: '...'
}
```

In general, if we omit the suffix, we can check the failures positionally:

```
failures: {
  0: '...',
  1: '...'
}
```

Which will be equivalent to:

```
failures: [
  '...',
  '...'
]
```

This way we don't need other characters and also don't extend the syntax.

lazarljubenovic:
I went for a different character because of cases like [this](https://github.com/mgechev/codelyzer/blob/accc942619f7c3e58db3ab2aabcc2ba31d2131dd/test/useHostPropertyDecoratorRule.spec.ts#L7-L11). How do we handle it?

PR #279, opened 2017-03-31
Automatic migration to <ng-template>

mgechev:


src/templateToNgTemplateRule.ts
```diff
@@ -0,0 +1,83 @@
...
+
+class TemplateToNgTemplateVisitor extends BasicTemplateAstVisitor {
+  private _prevClosing: number = 0;
+  private _fix: Fix;
+  private _visitedElements = new Set<EmbeddedTemplateAst>();
+
+  visitEmbeddedTemplate(element: EmbeddedTemplateAst, ctx: any) {
+    if (this._visitedElements.has(element)) {
+      return;
+    } else {
```
valorkin: You don't need 'else' if you did return inside of 'if' statement

src/angular/templates/templateParser.ts
```diff
@@ -69,12 +75,12 @@ export const parseTemplate = (template: string, directives: DirectiveDeclaration
 
   // Make sure it works with 2.2.x & 2.3.x
   const summaryKind = ((compiler as any).CompileSummaryKind || {}).Template;
-  const templateMetadata: compiler.CompileTemplateMetadata = {
+  let templateMetadata: any = {
```
valorkin: Using any is like a small step into dark side :)

mgechev: Agree. Trying to reduce it to minimum but the constant changes of the Angular's compiler API makes it impossible.

PR #286, opened 2017-04-02
fix(rules): order problem with multiple prefixes for selector

### LINKED ISSUE #275: Prefix order problem with multiple prefixes for selector
I noticed an issue with multiple selector prefixes. There seems to be an order dependency when one prefix contains another.

Example app component:
```typescript
import { Component } from '@angular/core';

@Component({
  selector: 'abc-app',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.css']
})

export class AppComponent {
  title = 'app works!';
}
```

tslint.json properties:
```json
    "directive-selector": [true, "attribute", ["ab", "abc"], "camelCase"],
    "component-selector": [true, "element", ["ab", "abc"], "kebab-case"],
```

With this setup, I get a TSLint error on the selector saying that it "should have one of the prefixes: ab,abc".

If I switch the selector to `ab-app`, there is no error. If I switch the prefix list to `["abc", "ab"]`, there is no error on either `ab-app` or `abc-app`.

I'm using an Angular 4 seed app with the following dependencies (plus more):
```json
    "@angular/core": "^4.0.0",
    "@angular/cli": "1.0.0",
    "codelyzer": "~2.0.0",
    "tslint": "~4.5.0",
    "typescript": "~2.2.0"
```

leosvelperez:
Closes #275.

### DIFF
--- a/src/selectorNameBase.ts
+++ b/src/selectorNameBase.ts
@@ -85,6 +85,11 @@ export abstract class SelectorRule extends Lint.Rules.AbstractRule {
   }
 
   private setPrefixValidator(prefix: any, name: string) {
+    if (!this.isMultiPrefix) {
+      prefix = (prefix||[]).sort((a: string, b: string) => {
+        return a.length < b.length ? 1 : -1;
+      });
+    }
     let prefixExpression: string = this.isMultiPrefix?prefix:(prefix||[]).join('|');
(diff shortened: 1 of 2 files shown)


mgechev:
Thanks for the PR!

sheridjs:
Thanks!

leosvelperez:
Glad to help!

PR #312, opened 2017-05-31
doc: component-selector: allowing element&attribute

karol-depka:


mgechev:
Thanks for the PR!

karol-depka:
@mgechev Thanks to you! For completeness, the schema there should also be updated most probably: http://codelyzer.com/rules/component-selector/ . But for now I wanted to avoid a rabbit hole :). Btw, do you happen to come to Barcelona Angular Camp, Jul 6-7?

mgechev:
No, unfortunately I'll miss it. Trying to reduce my traveling recently 😞 .

PR #317, opened 2017-06-08
feat: add a no-negated-async rule

connor4312:
Angular's async pipes emit `null` initially, prior to the observable emitting
any values, or the promise resolving. This can cause negations, like
`*ngIf="!(myConditional | async)"` to thrash the layout and cause expensive
side-effects like firing off XHR requests for a component which should not
be shown.

This PR adds a linting rule which fails on usages of !(expr | async) and
(expr | async) == false.


connor4312:
It looks like the failing build is an issue with node-sass on Travis' 32-bit slaves. Tests pass otherwise here!

mgechev:
@connor4312 thanks for the PR, looks great! I will review and merge as soon as possible.

Looks great, at first!

mgechev:
@connor4312 merged. Great job!

connor4312:
Thank you!

PR #340, opened 2017-06-20
fix: handle case when whitespace on one side of interpolation

### LINKED ISSUE #335: Angular Whitespace "check-interpolation" fails
...with `{{ foo}}` and `{{foo }}`.

mgechev:
Fix #335


wKoza:
It's not directly the subject but can we add a assertTrue for the pipe like this : 
```js 
  it('should work when space before and after "|"', () => {
      let source = `
      @Component({
        selector: 'foo',
        template: \`
          <div>{{ foo | async }}</div>
        \`
      })
      class Bar {
        foo: any;
      }
      `;
      assertSuccess('angular-whitespace', source);
    });
```

In a future release, we can manage the chain of pipes like this `{{ foo | async |      bar }}` ?

PR #357, opened 2017-06-25
Improve tslint rules

rokerkony:
Hey :)
Each rule change is in separate commit. I can add other rules as well but I guess step by step is better approach.

One of commits is adding `npm run lint:fix` command which is using built-in fixers and makes it easy to solve development issues.

I'm using locally even pre-commit hook to run this fixer on each commit.

**.git/hooks/pre-commit**
```sh
#!/bin/sh
FILES=`git status --porcelain -uno | grep \.ts$`
FILES+="EMPTY"

echo "Modified files:"
git status --porcelain -uno

if [ "$FILES" != "EMPTY" ]; then
   echo "There are typescript files to be linted. Processing . . ."
   if [[ $(npm run lint:fix | grep "error(s)") ]]; then
       echo "Fixer found some coding standard errors and fixed all for you, but commit was stopped, please commit again!"
       exit 1
   fi
fi
```


tslint.json
```diff
@@ -25,6 +39,18 @@
...
+    "whitespace": [
+      true,
+      "check-branch",
+      "check-decl",
+      "check-module",
+      "check-operator",
+      "check-preblock",
+      "check-separator",
+      "check-type",
+      "check-typecast"
```
mgechev: Let's disable this, it looks more clear what the target is when there is no space after type casting.

src/angular/ngWalker.ts
```diff
@@ -195,7 +195,7 @@ export class NgWalker extends Lint.RuleWalker {
     }
     const sf = ts.createSourceFile(path, `\`${content}\``, ts.ScriptTarget.ES5);
     const original = sf.getFullText;
-    sf.getFullText = function () {
+    sf.getFullText = function() {
```
mgechev: I am usually for white spaces after `function` keyword in anonymous functions. I don't have any specific reasoning, so if you have an argument against it, I am fine :-)

tslint.json
rokerkony: Absolutely agree, removed.
Only advantage to have it in place is to force consistency. It would be probably better to make a feature request to tslint and have a possibility to decide about space for each option separately.

src/angular/ngWalker.ts
rokerkony: I have this rule set to `always` in my own projects but it would require space before parens also in methods. E.g.: `export function assertFailures<SPACE_REQUIRED>(ruleName: string, source: string, fails: IExpectedFailure[], options = null) {`

Specific reasoning is only to have a consistency. So either I will remove a rule or I will change it to `always` and apply the rule on all functions ...

mgechev: Ok, we can keep it.

Issue #346, opened 2017-06-21
check-pipe breaks with ngFor


cexbrayat:
It looks like the new `check-pipe` rule breaks with ngFor (as of 3.1.1):

    <div *ngFor="let user of users | slice:0:4">

Throws:

    The pipe operator should be surrounded by one space on each side, i.e. " | "

See #347 for a unit test reproducing the issue.

intellix:
and an ngSwitch at least. Not sure which others are missing

mgechev:
Thanks for reporting the issue. I will take a look during the weekend on AngularUP.

rokerkony:
Other cases which fail on the `check-pipe` rule:

1.  `{{ (foo) | pipe }}` our code example: `{{ (totalPrice || price.total) | eur }}`
2. `<div *ngIf="items$ | async as items; else loading;"></div>`

rokerkony:
@mgechev this issue is closed but the use case n.2 was not fixed... can you take a look on it?

mgechev:
There's another issue related to `"check-pipe"` https://github.com/mgechev/codelyzer/issues/365. Looks the same as the one reported by you.

rokerkony:
ok, thx :)

cexbrayat:
@mgechev I tried 3.1.2 but it still fail for `ngFor` in a simple use case, with an external template.

See #368 for a unit test reproducing the issue.

mgechev:
See https://github.com/mgechev/codelyzer/issues/365

cexbrayat:
@mgechev Ha sorry, I thought #365 was unrelated to this one (the title was not really explicit :) ).

mgechev:
No worries

PR #368, opened 2017-07-05
test: add failing test for check-pipe in ngFor with external template

cexbrayat:
Adds a failing test to reproduce [#346](https://github.com/mgechev/codelyzer/issues/346 "check-pipe breaks with ngFor") with an external template.


wKoza:
Closed via #401

PR #370, opened 2017-07-09
Decorator not allowed

### LINKED ISSUE #363: Compiler should throw an error if I use @HostListener in a service
> I'm submitting a ...
> 
> 
> [ ] Regression (behavior that used to work and stopped working in a new release)
> [ ] Bug report 
> [ X ] Feature request
> [ ] Documentation issue or request
> [ ] Support request => Please do not submit support request here, instead see https://github.com/angular/angular/blob/master/CONTRIBUTING.md#question
> Current behavior
> 
> I'm allowed to put @HostListener in a service, even if it doesn't work
> 
> Expected behavior
> 
> The compiler should fail (or at least emit a warning?) telling me that the use of @HostListener within a service (or any other place that doesn't have a host) won't work.
> 
> Minimal reproduction of the problem with instructions
> 
> What is the motivation / use case for changing the behavior?
> 
> Just a quality-of-life improvement

By @adamdport 
Source https://github.com/angular/angular/issues/17847

wKoza:
closes #363

### DIFF
--- a/src/angular/ngWalker.ts
+++ b/src/angular/ngWalker.ts
@@ -98,6 +98,12 @@ export class NgWalker extends Lint.RuleWalker {
 
   protected visitClassDecorator(decorator: ts.Decorator) {
     let name = getDecoratorName(decorator);
+
+    if (name === 'Injectable') {
+      console.log('test');
+      this.visitNgInjectable(<ts.ClassDeclaration>decorator.parent, decorator);
+    }
+
     // Not invoked @Component or @Pipe, or @Directive
     if (!(<ts.CallExpression>decorator.expression).arguments ||
         !(<ts.CallExpression>decorator.expression).arguments.length ||
@@ -146,6 +152,8 @@ export class NgWalker extends Lint.RuleWalker {
   protected visitNgDirective(metadata: DirectiveMetadata) {}
 
   protected visitNgPipe(controller: ts.ClassDeclaration, decorator: ts.Decorator) {}
+
+  protected visitNgInjectable(controller: ts.ClassDeclaration, decorator: ts.Decorator) {}
 
   protected visitNgInput(property: ts.PropertyDeclaration, input: ts.Decorator, args: string[]) {}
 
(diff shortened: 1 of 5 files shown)


src/angular/ngWalker.ts
```diff
@@ -34,9 +38,9 @@ export interface NgWalkerConfig {
 
 export class NgWalker extends Lint.RuleWalker {
   constructor(sourceFile: ts.SourceFile,
-    protected _originalOptions: Lint.IOptions,
-    private _config?: NgWalkerConfig,
-    protected _metadataReader?: MetadataReader) {
+              protected _originalOptions: Lint.IOptions,
```
mgechev: Would you change back to spaces

src/decoratorNotAllowedRule.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  className: string;
+  isInjectable = false;
+
+  constructor(sourceFile: ts.SourceFile, private rule: Rule) {
+    super(sourceFile, rule.getOptions());
+  }
+
+  visitNgInjectable(controller: ts.ClassDeclaration, decorator: ts.Decorator) {
+    this.className = controller.name.text;
```
mgechev: In case of injectable (i.e. service) it's not necessary a controller.

src/angular/ngWalker.ts
mgechev: It seems it's spaces already. You can ignore this one.

mgechev:
LGTM. There is only one minor comment. I will take more detailed look tonight.

You are reporting errors only in case of `@Injectable` to prevent false negatives for base component / directive classes?

src/decoratorNotAllowedRule.ts
wKoza: What do you suggest?

mgechev: Just a rename `controller` -> `classDeclaration`.

wKoza:
This rule verify that each class decorator uses its decorators allowed. For this moment, I'm reporting errors only in case of `@Injectable`.

rafaelss95:
@wKoza do you have plans to extend this rule to other decorators, like `@Pipe` and others?

wKoza:
Yes @rafaelss95, we can extend this to `@Pipe` et `@NgModule`. Do you have another ideas ?

Issue #365, opened 2017-07-04
check-pipe totally broken


Martin-Wegner:
Since 3.1.2 check-pipe is totally broken.

```
<div *ngIf="countingPoints$ | async as countingPoints; else countingPoinsAreLoading">
  {{countingPoints}}
</div>
<ng-template #countingPoinsAreLoading>loading</ng-template>
```

> ERROR: XXX.component.html[2, 13]: The pipe operator should be surrounded by one space on each side, i.e. " | ".

mgechev:
Thanks for reporting the issue. There are two issues in this case:

- The warning is reported for code having a valid style.
- The warning is reported on incorrect location.

I'll push a fix in upcoming releases.

willgm:
Similar error here:

> ERROR: /home/willgm/proj/jexia-webapp/src/app/project/project-features/project-features.component.html[1, 25]: The property "async" that you're trying to access does not exist in the class declaration.

For this line:
```html
<div *ngIf="(currentProject | async).ui.features as features">
```

mgechev:
@Martin-Wegner not a related issue. Please, take a look at the search.

willgm:
@mgechev, sorry, but I can't see how this error is related to this issue. Can you explain?

I searched a little more, looks like there are other issues more likely to math the errors above: #190 and angular/angular-cli/issues/6554

This issue appeared after I updated from `3.0.1` to `3.1.0` or `3.0.2`:
```
~/proj/jexia-webapp [work ~2 *] » npm i -D codelyzer@3.1.0                              willgm@localhost
+ codelyzer@3.1.0
updated 1 package in 4.362s
---------------------------------------------------------------------
~/proj/jexia-webapp [work ~2 *] » npm run lint                                          willgm@localhost

> web-app@0.0.3-alpha lint /home/willgm/proj/jexia-webapp
> ng lint --type-check


ERROR: /home/willgm/proj/jexia-webapp/src/app/project/project-features/project-features.component.html[1, 2]: The property "async" that you're trying to access does not exist in the class declaration.
ERROR: /home/willgm/proj/jexia-webapp/src/app/project/project-features/project-features.component.html[55, 34]: The property "async" that you're trying to access does not exist in the class declaration.

Lint errors found in the listed files.
npm ERR! code ELIFECYCLE
npm ERR! errno 2
npm ERR! web-app@0.0.3-alpha lint: `ng lint --type-check`
npm ERR! Exit status 2
npm ERR! 
[...]

PR #385, opened 2017-07-25
Demonstrate bug with check-pipe

dhalperi:
Why does changing the variable name in the `let` statement cause `check-pipe` to fail?


dhalperi:
I guess this is related to #365

Issue #377, opened 2017-07-13
i18n best practices


mgechev:
Consider to add the following i18n best practice:

> ## Set a custom id to improve search and maintenance
> 
> The angular i18n extractor tool generates a file with a translation unit entry for each i18n attribute in a template. By default, it assigns each translation unit a unique id such as this one:
> 
>  content_copy
> <trans-unit id="ba0cc104d3d69bf669f97b8d96a4c5d8d9559aa3" datatype="html">
> This id is obscure and difficult for humans to read or remember.
> 
> Worse, when you change the translatable text, perhaps to fix a typo, the extractor tool generates a new id for that translation. You will lose the translation unless you update it with the new id. That complicates maintenance.

- [x] - Look for elements without custom ID.
- [x] - Look for elements with text content but no i18n attribute.

https://angular.io/guide/i18n#set-a-custom-id-to-improve-search-and-maintenance

abierbaum:
Would it be possible to also add a rule that looks for elements in component templates that have text content but no i18n attribute?

One issue we are having on our team is simply enforcing that people mark all the exposed strings for extraction and translation.

mgechev:
This should be relatively straightforward. I will add it as part of this issue.

abierbaum:
@mgechev Awesome.  Thanks!!!

mgechev:
@abierbaum both rules are implemented. Should be ready to use in codelyzer@3.2.0.

panuruj:
In general, I would not recommend overriding the auto-generated ID. In a project that have thousands of strings to be translated, it is error prone to rely on the developer to ensure that the ID is unique across the entire application.

abierbaum:
@mgechev Thanks for adding this!!!

PR #386, opened 2017-08-02
Implement i18n rules from #377

mgechev:
Fix https://github.com/mgechev/codelyzer/issues/377


test/i18nRule.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+        },
+        ['check-text']
+      );
+    });
+
+    it('should fail with missing id string', () => {
+      let source = `
+      @Component({
+        template: \`
+          <div>Text {{ foo }}</div>
```
wKoza: can you also add a test with `<div i18n>{{text}} test</div>` which should fail.

src/noAccessMissingMemberRule.ts
```diff
@@ -98,19 +111,21 @@ class SymbolAccessValidator extends RecursiveAngularExpressionVisitor {
...
+      })
+      .sort((a, b) => a.distance - b.distance);
     const first = tmp.shift();
     if (!first) {
       return [];
     } else {
       result.push(first);
       let current: any;
-      while (current = tmp.shift()) {
+      while ((current = tmp.shift())) {
```
mgechev: Yes, prettier has set it.

mgechev:
@wKoza thanks for the review. Addressed your comments.

Issue #394, opened 2017-08-16
no-access-missing-member issue with ngForm template variable


csvn:
## Expected behavior
No errors

## Actual behavior
I get the error `The property "form" that you're trying to access does not exist in the class declaration.`

## Minimum reproducible code
```typescript
// src/app.component.ts
import { Component } from '@angular/core';

@Component({
  selector: 'my-app',
  template: `
    <div ngForm #form="ngForm">
      Valid: {{ form.valid }}
    </div>
  `
})
export class AppComponent {}
```

```json
// tslint.json
{
  "rulesDirectory": [
    "node_modules/codelyzer"
  ],
  "rules": {
    "no-access-missing-member": true
  }
}
```

```json
// tsconfig.json
{
  "compilerOptions": {
    "moduleResolution": "node",
    "experimentalDecorators": true
  },
  "include": [
    "src/**/*"
  ]
}
```

## Suggestion
[Change this line's](https://github.com/mgechev/codelyzer/blob/master/src/angular/config.ts#L66) selector from `'form'` to `'form, [ngForm]'`. This will match the definition better for Angular's [NgFormDirective](https://angular.io/api/forms/NgForm#selectors).

### Versions
Angular: v4.3.4
Tslint: v5.6.0
Codelyzer: v3.1.2

csvn:
I created a pull request if this seems reasonable to the maintainers. #396

PR #396, opened 2017-08-16
fix: support ngForm for directives

csvn:
See https://github.com/mgechev/codelyzer/issues/394

### DIFF
--- a/src/angular/config.ts
+++ b/src/angular/config.ts
@@ -63,7 +63,7 @@ export const Config: Config = {
   },
 
   predefinedDirectives: [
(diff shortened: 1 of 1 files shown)


wKoza:
@csvn , thanks for the PR ! Why not also add `form:not([ngNoForm]):not([formGroup])`  ?

csvn:
@wKoza That might create errors in existing code, but that might be better as it is more correct. I can add it for sure, I just started the PR a bit conservative.

mgechev:
Thanks for the PR!

PR #401, opened 2017-09-03
Fix desugared directive

### LINKED ISSUE #379: some rules report false positives
I just discovered some weird cases where `angular-whitespace` complains about:
- `<div fxLayout="column">` (the `l` - as in the letter `L` - must be the confusing part)
- `<md-icon *ngIf="isOnline | async; else offline">cloud_off</md-icon>`
- `<md-toolbar color="primary">` - might be the same case as the first one
- `<rj-image-filters (queryChange)="updateFilters($event)"></rj-image-filters>` - could be the same case with the letter `L`
- `<!-- Setttings -->` - might be the `!` sign there, nonetheless, comments should be ignored
- `<md-select [(ngModel)]="timeRangeKey">` - could be the `L` again
- `*ngIf="query.isUpdating | async as isUpdating"` - there's obviously space there, so not sure what the problem is

And `banana-in-a-box`:
- `<a md-list-item (click)="navigate(['/resources'])">` - confuses what's inside the quotes I guess

Finally `no-access-missing-member `:
- complains about a prop not existing on the class, but it's actually a local var
```html
<rj-paginated-resources *ngIf="query"
							#pagination="rjPagination"
							[type]="type"
							[query]="query | async">
	<div fxLayout="row"
		 fxLayoutWrap
		 class="images"
		 padding>
		<ng-container *ngFor="let image of pagination.items | async; trackBy: trackByImage">
			<div class="image">
				<rj-image-card [image]="image"></rj-image-card>
			</div>
		</ng-container>
	</div>
</rj-paginated-resources>
```

wKoza:
Closes #379 

This solution is not graceful but it allows to recover a higher-level parent of the AST in our visitor. I think that this information was generally lacking. I did not find any other way to do it. 

Another problem: We lose the correct place where the source code doesn't match the rule (see templatesUsePublicRule.spec.ts) in our tests. In my IDE (Webstrom), there is no position error.


mgechev:
Awesome job! Will take a look as soon as possible.

mgechev:
@wKoza there might be an issue in the testing package implementation.

src/angularWhitespaceRule.ts
```diff
@@ -152,11 +152,21 @@ class WhitespaceTemplateVisitor extends BasicTemplateAstVisitor {
...
 class PipeWhitespaceVisitor extends RecursiveAngularExpressionVisitor implements ConfigurableVisitor {
   visitPipe(ast: ast.BindingPipe, context: RecursiveAngularExpressionVisitor): any {
-    const start = ast.span.start;
-    const exprStart = context.getSourcePosition(ast.exp.span.start);
-    const exprEnd = context.getSourcePosition(ast.exp.span.end);
-    const sf = context.getSourceFile().getFullText();
-    const exprText = sf.substring(exprStart, exprEnd);
+
+    let exprStart, exprEnd, exprText, sf;
+    if (NgWalker.prop && ( NgWalker.prop.templateName === 'ngForOf' || NgWalker.prop.templateName === 'ngIf')) {
```
mgechev: Maybe we can preserve the context (`prop`) from a parent element visitor?

mgechev: Isn't this valid for all different templates? `*` attribute prefix always desugars to `<ng-template...></ng-template>`.

wKoza: Yes, a test on `prop.templateName` should be enough.

wKoza: Yes, this is an evil way. After a good night, now, we have our context the parent class `sourceMappingVistor`. It's better.

wKoza: Finally, I modified `SourceMappingVisitor.getSourcePosition()` method, which is more central. So, this fix the position problem in our tests.

Issue #437, opened 2017-10-23
Built-in directive config not compatible with material2 beta.12


csvn:
All `md-/Md` prefixes has been removed in favor of `mat-/Mat` prefix. To those who upgrades, the codelyzer rule `no-access-missing-members` now throws errors on valid code.

```
<button [matMenuTriggerFor]="menu">menu</button>
<mat-menu #menu="matMenu">
</mat-menu>
```
> "The property "menu" that you're trying to access does not exist in the class declaration.

The cause for this is that the configuration only exports `mdMenu`, not `matMenu`. Is it possible to have multiple exports (e.g. `exportAs: 'mdMenu, matMenu'`)? There are also several directives that are not configured correctly for the `mat-` prefix.
https://github.com/mgechev/codelyzer/blob/master/src/angular/config.ts#L69-L74

I'd be happy to make a pull request, but I'm not sure what to do for backwards compatibility to support both options. Maybe it's best to just split `md` & `mat` into their own like below?

```
// material2 deprecated `md` prefixes
{ selector: 'md-button-toggle-group', exportAs: 'mdButtonToggleGroup' },
{ selector: 'md-select', exportAs: 'mdSelect' },
{ selector: 'md-menu', exportAs: 'mdMenu' },
{ selector: '[md-menu-item]', exportAs: 'mdMenuItem' },
{ selector: '[md-menu-trigger-for], [mdMenuTriggerFor]', exportAs: 'mdMenuTrigger' },
{ selector: '[md-tooltip], [mdTooltip]', exportAs: 'mdTooltip' },

// material2 official `mat` prefixes
{ selector: 'mat-button-toggle-group', exportAs: 'matButtonToggleGroup' },
{ selector: 'mat-select', exportAs: 'matSelect' },
{ selector: 'mat-menu', exportAs: 'matMenu' },
{ selector: '[mat-menu-item]', exportAs: 'matMenuItem' },
{ selector: '[mat-menu-trigger-for], [matMenuTriggerFor]', exportAs: 'matMenuTrigger' },
{ selector: '[mat-tooltip], [matTooltip]', exportAs: 'matTooltip' },
```

In #395 I was told that `no-access-missing-members` would be removed, but I have seen no evidence of this yet in the v4 beta branch. Right now it's pretty messy, where I have to copy all the settings and adjust them. Any info about this would be appreciated.

mgechev:
@csvn your suggestion sounds good. Would you open a PR with the update?

[...]

PR #438, opened 2017-10-23
fix: directive config for material2 beta.12

csvn:
Related to #437

I noticed that there seem to be a lot of directives that has exports that are not configured (e.g. `mat-button` with `matButton` export). But I deemed it out of scope here, I just added the equivalent `mat` selectors and exports for the existing directives.

I also had trouble running tests on windows (there is no `cp` command), so I'll just hope it passes on the CI.


mgechev:
Awesome, thanks for the PR! Today I'll issue a new release with exported i18n rules and your change.

Issue #412, opened 2017-09-12
Issue in `check-pipe`


mgechev:
By @rolandjitsu

> I'm still seeing some problems with the `check-pipe` rule for:
> ```html
> <p *ngIf="(items | async)?.length as length" class="mat-caption">
> ```
> 
> ```html
> <span *ngIf="product.propagationTolerance" class="label">&plusmn; {{ product.propagationTolerance | number }}</span>
> ```
> 
> For the latter, removing `&plusmn;` seems to fix it.

wKoza:
@mgechev , do you have any idea for manage html entities (we have a problem with the position) ?

mgechev:
@wKoza, the compiler converts them to unicode chars?

wKoza:
yes, @mgechev

mgechev:
Not sure if there's a way to handle this, except the one I shown above, used by the language service. However, this way, I'm afraid we're getting coupled even further to lower, unstable APIs :-/.

wKoza:
BTW, this issue (with html entity) happens  even with check-semicolon :(

mgechev:
https://github.com/mgechev/codelyzer/issues/460

PR #448, opened 2017-10-29
 refactor: drop rules

### LINKED ISSUE #264: Drop rules which produce warnings that are supposed to be compile-time errors
Drop:

- `templates-use-public`
- `no-access-missing-member`
- `invoke-injectable`

mgechev:
Fix #264


README.md
```diff
@@ -62,7 +62,7 @@ Create the following `tslint.json` file like:
 {
   "extends": [ "codelyzer" ],
   "rules":{
-    "angular-whitespace": [true, "check-interpolation", "check-pipe", "check-semicolon"],
```
wKoza: Why are you drop the `check-pipe` rule ? In relation with #412 ?

mgechev: Yes, for now it's dropped only from the readme until we figure it out :-)

PR #456, opened 2017-11-11
[feat] add trackByFunctionRule

### LINKED ISSUE #378: [feat] add a warning when not using trackBy with ngFor
Even though its considered best practise, I always forget to add a trackBy when using ngFor. It would be awesome to have a rule to enforce this in templates. Since invalid usage of trackBy is currently swallowed silently by angular, as a stretch it would be cool if the rule validates the trackBy expression as well. If you're in favour of this I would love to take a go at a PR for this feature. Thanks! 😄

wKoza:
closes #378


src/trackByFunctionRule.ts
```diff
@@ -0,0 +1,69 @@
...
+  public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+    return this.applyWithWalker(
+      new NgWalker(sourceFile,
+        this.getOptions(), {
+          templateVisitorCtrl: TrackByTemplateVisitor,
+        }));
+  }
+}
+
+const noTrackByRe = new RegExp('trackBy:');
```
mgechev: `/trackBy\s*:/` might be a better regex because of the whitespace which is ignored by default

mgechev:
@wKoza good job! Left only one comment.

mattlewis92:
Awesome, thanks for your work on this @wKoza! 👍

Issue #469, opened 2017-12-08
two bugs of angular-whitespace


sagittarius-rev:
**1. improper position of the errors with angular-whitespace**

For example,  the template is:

```html
<div *ngFor="let item of list;trackBy item.id;let $index=index">{{ item.title }}</div>
```

When lint this Angular template, it would locate an error before `trackBy` (Missing white space after semicolon) as expected. But actually, it will locate the error before ` item.id` .

This bug caused by `angularWhitespaceRule.ts` of line 86, line 97 and line 105:

`directive.split('=')[1].trim()`

It supposes that the right side code of the directive would equal to `directive.split('=')[1]`, but the microsyntax of the Angular directives allow equality sign existed within the expression! In this example, the rawExpression would be `let item of list;trackBy item.id;let $inde` wrongly, instead of `let item of list;trackBy item.id;let $index=index` .

**How to fix this bug?**

first, replace the code of line 86

```typescript
      const rawExpression = directive.split('=')[1].trim();
```

with

```typescript
      const match = /^([^=]+=\s*)([^]*?)\s*$/.exec(directive);
      const rawExpression = match[2];
      const positionFix = match[1].length + 1;
```

second, replace the code of line 97 and line 105

```typescript
const start = prop.sourceSpan.start.offset + internalStart + directive.length - directive.split('=')[1].trim().length + 1;
```

with

```typescript
const start = prop.sourceSpan.start.offset + internalStart + positionFix;
```

**2. mistake of the auto fix with angular-whitespace**

We can use `tslint --fix` to auto fix the errors of Angular template, for example:

```html
<h4>{{title|translate    }}</h4>
```

As expected, the fix result would be:

```html
<h4>{{ title | translate }}</h4>
```

But actually, the result is:

```html
<h4>{{ titl e |translate }}</h4>
```

This bug caused by `angularWhitespaceRule.ts` of the code between line 48 and line 56:

```typescript
          context.addFailure(
            context.createFailure(
            start, length, failure, [
            new Lint.Replacement(
              absolutePosition,
              length,
              `${InterpolationOpen} ${match[0].replace(InterpolationOpen, '').replace(InterpolationClose, '').trim()} ${InterpolationClose}`
            )
          ]));
```

Normally, when locating and fixing the errors, it must minimize the changes of the text, to avoid disturbing other fix actions. But the codes above changes the whole text of Angular interpolation, including `{{` and `}}`.

[...]

PR #470, opened 2017-12-08
fix two bugs of angular-whitespace, update the spec code

sagittarius-rev:
See issue #469


wKoza:
Awesome @sagittarius-rev ! Before merging your branch, can you add a test with this case (good positon) :

`<div *ngFor="let item of list;trackBy item.id;let $index=index">{{ item.title }}</div>`

And your test (fix) with an Interpolation and pipe : 

`<h4>{{title|translate    }}</h4>`

sagittarius-rev:
add two test cases

PR #471, opened 2017-12-11
check multiple semicolons inside the directive expressions

sagittarius-rev:
**check multiple semicolons inside the directive expressions for angular-whitespace rule**

The `check-semicolon` rule of `angular-whitespace` could only check out and fix the first error when there are mutiple semicolon errors inside the directive expressions. For example:

```html
  <h3 *ngIf="date | date:'fullDate';let dateString;else errorDate">{{ dateString }}</h3>
  <ng-template #errorDate>error date!</ng-template>
```

Only the first error ( `;let` ) should be check out and fixed. If someone want to fix all of two errors of this example, it must run the `tslint --fix` command twice.

I have rewritten the codes of `check-semicolon` to solve the problem, also add 5 test cases, one for success check, and 4 for fail/apply proper replacements.


mgechev:
https://github.com/mgechev/codelyzer/issues/472

Issue #473, opened 2017-12-17
Feature request: no output named after a standard DOM event


gbilodeau:
Naming a custom directive output after a standard DOM event can cause unwanted behaviour because of event bubbling. As explained clearly [here](https://stackoverflow.com/questions/40276400/what-are-recommanditions-for-output-event-names-to-prevent-native-event-name-c), declaring a listener for a custom directive output with a "submit" name will be invoked not only when this output emits a new value but also when a native "submit" event bubbles up from a possible descendant form.

This issue has also been discussed [here](https://github.com/angular/angular/issues/13997), resulting in a statement that this behaviour is just "common sense".  Since common sense isn't necessarily obvious to new users of the framework, it would be useful to create a rule that highlights such a naming practice error.

There are no guidelines with regards to this in the Angular style guide.

PR will follow.

PR #474, opened 2017-12-17
No output named after standard event

gbilodeau:
PR implementing #473 
Event list was taken from https://developer.mozilla.org/en-US/docs/Web/Events

This is my first time submitting a pull request anywhere, apologies in advance for any errors!


src/noOutputNamedAfterStandardEventRule.ts
```diff
@@ -0,0 +1,66 @@
...
+  public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+    return this.applyWithWalker(
+      new OutputMetadataWalker(sourceFile,
+        this.getOptions()));
+  }
+}
+
+
+export class OutputMetadataWalker extends NgWalker {
+  private readonly standardEventNames = [
```
mgechev: Let's use hash map here for constant check.

PR #478, opened 2017-12-31
feat(rule): add no-template-call-expression rule

kevinphelps:
This is a rule that I wrote for an internal project. I thought it might be useful to others.

### DIFF
--- a/src/noTemplateCallExpressionRule.ts
+++ b/src/noTemplateCallExpressionRule.ts
@@ -1,0 +1,48 @@
+import * as Lint from 'tslint';
+import * as ts from 'typescript';
+import { NgWalker, NgWalkerConfig } from './angular/ngWalker';
+import { BasicTemplateAstVisitor  } from './angular/templates/basicTemplateAstVisitor';
+import { RecursiveAngularExpressionVisitor  } from './angular/templates/recursiveAngularExpressionVisitor';
+import * as e from '@angular/compiler/src/expression_parser/ast';
+
+export class Rule extends Lint.Rules.AbstractRule {
+  public static metadata: Lint.IRuleMetadata = {
+    ruleName: 'no-template-call-expression',
+    type: 'functionality',
+    description: 'Call expressions are not allowed in templates except in output handlers.',
+    rationale: 'The change detector will call functions used in templates very often. Use an observable instead.',
+    options: null,
+    optionsDescription: 'Not configurable.',
+    typescriptOnly: true,
+  };
+
+  static FAILURE_STRING = 'Call expressions are not allowed in templates except in output handlers.';
+
+  apply(sourceFile: ts.SourceFile) {
+    const walkerConfig: NgWalkerConfig = {
+      templateVisitorCtrl: TemplateVisitor,
+      expressionVisitorCtrl: ExpressionVisitor
+    };
+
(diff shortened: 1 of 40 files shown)


mgechev:
@kevinphelps awesome job! I will take a look later today/tomorrow :-).

src/noTemplateCallExpressionRule.ts
```diff
@@ -0,0 +1,48 @@
...
+    ruleName: 'no-template-call-expression',
+    type: 'functionality',
+    description: `Call expressions are not allowed in templates except in event handler bindings.`,
+    rationale: `The change detector will call functions used in templates very often. Use an observable instead.`,
+    options: null,
+    optionsDescription: `Not configurable.`,
+    typescriptOnly: true,
+  };
+
+  static FAILURE_STRING = 'Call expressions are not allowed in templates except in event handler bindings.';
```
mgechev: Let's change this to:

>Call expressions are not allowed in templates except in output handlers.

The current one is correct as well, this way we're more consistent with the Angular's terminology.

src/noTemplateCallExpressionRule.ts
```diff
@@ -0,0 +1,48 @@
...
+import { BasicTemplateAstVisitor  } from './angular/templates/basicTemplateAstVisitor';
+import { RecursiveAngularExpressionVisitor  } from './angular/templates/recursiveAngularExpressionVisitor';
+import * as e from '@angular/compiler/src/expression_parser/ast';
+
+export class Rule extends Lint.Rules.AbstractRule {
+  public static metadata: Lint.IRuleMetadata = {
+    ruleName: 'no-template-call-expression',
+    type: 'functionality',
+    description: `Call expressions are not allowed in templates except in event handler bindings.`,
+    rationale: `The change detector will call functions used in templates very often. Use an observable instead.`,
```
mgechev: We can use a string literal here instead of template strings. There are already existing rules where this should be fixed as well.

src/noTemplateCallExpressionRule.ts
```diff
@@ -0,0 +1,48 @@
...
+    options: null,
+    optionsDescription: `Not configurable.`,
```
mgechev: Same here.

src/noTemplateCallExpressionRule.ts
```diff
@@ -0,0 +1,48 @@
...
+import { NgWalker, NgWalkerConfig } from './angular/ngWalker';
+import { BasicTemplateAstVisitor  } from './angular/templates/basicTemplateAstVisitor';
+import { RecursiveAngularExpressionVisitor  } from './angular/templates/recursiveAngularExpressionVisitor';
+import * as e from '@angular/compiler/src/expression_parser/ast';
+
+export class Rule extends Lint.Rules.AbstractRule {
+  public static metadata: Lint.IRuleMetadata = {
+    ruleName: 'no-template-call-expression',
+    type: 'functionality',
+    description: `Call expressions are not allowed in templates except in event handler bindings.`,
```
mgechev: We can use a string literal here instead of template strings. There are already existing rules where this should be fixed as well.

kevinphelps:
I replaced all template literals with string literals where possible.

mgechev:
Thank you for the extra help!

PR #482, opened 2018-01-08
fix(core): update the CSS parser

### LINKED ISSUE #446: Codelyzer hanging for "styles" attribute in Angular component
I am developing an application using @angular/cli 1.4.9 and @angular 4.4.x , but have encountered an issue running `ng lint` on the code recently.  I've diagnosed this using ` tslint -c tslint.json .\src\some.component.ts` which never returns.  

I have traced this down to the following code snipet:

```typescript
@Component({
  selector: 'div[some-component]',
template: `<ng-content></ng-content>`,
  styles: [
  `:host {
    display:        flex;
    flex:           1 100%;
    margin:         20px;"
  }`
  ],
  changeDetection: ChangeDetectionStrategy.OnPush
})
```

Specifically the _styles_ options, if I remove this the above command succeeds.

I have played with different options in the tslint.json file, but find that many of them cause this not a specific one.

Is there anyway I can get diagnostic logging out to assist in this?

mgechev:
Fix #446

### DIFF
--- a/src/angular/styles/chars.ts
+++ b/src/angular/styles/chars.ts
@@ -1,3 +1,5 @@
+/* tslint:disable */
+
 /**
  * @license
  * Copyright Google Inc. All Rights Reserved.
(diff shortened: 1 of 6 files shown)


wKoza:
Good refactoring ! this parser needed it ;)
can you add a test to check that parsing no longer blocks ?

mgechev:
@wKoza good catch...the issue still persists in the parser. Trying to figure out where exactly.

mgechev:
@wKoza push the fix upstream as well.

mgechev:
`CssParser` is a module which Angular doesn't use internally. It's not part of the compiler bundle. We might be the only ones using it.

PR #483, opened 2018-01-09
fix(trackBy): fix rule when ngFor statements include let assignments

alisd23:
- No longer split by "=" as this broke `*ngFor` statements with `let i = index` statements or similar.
- Add test to cover cases like this.
- Fix line length lint error (maybe only on windows).

### DIFF
--- a/src/trackByFunctionRule.ts
+++ b/src/trackByFunctionRule.ts
@@ -24,7 +24,7 @@ export class Rule extends Lint.Rules.AbstractRule {
   }
 }
 
-const ngForExpressionRe = new RegExp(/\*ngFor="(.+)"/);
+const ngForExpressionRe = new RegExp(/\*ngFor\s*=\s*(?:'|")(.+)(?:'|")/);
 const trackByRe = new RegExp(/trackBy\s*:/);
 
 class TrackByNgForTemplateVisitor extends BasicTemplateAstVisitor {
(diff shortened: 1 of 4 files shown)


alisd23:
I also upgraded node-sass on my side because `npm install` would not work for me with that version as it couldn't find the binding - I think it doesn't exist for modern versions of node (but I'm not 100% sure of the exact issue). Let me know if you want me to revert that upgrade, and I can just use the CI to see if the tests are passing.

src/trackByFunctionRule.ts
```diff
@@ -24,28 +24,26 @@ export class Rule extends Lint.Rules.AbstractRule {
   }
 }
 
+const ngForExpressionRe = new RegExp(/\*ngFor="(.+)"/);
```
mgechev: `/\*ngFor\s*=\s*'|"(.+)'|"/` this might be more accurate it will catch`*ngFor = 'foobar'` as well.

mgechev:
@alisd23 thanks for the PR! Left one comment regarding the regexp.

src/trackByFunctionRule.ts
alisd23: @mgechev Yep, good point. Will add that in. Also did you see my comment above about node-sass?

mgechev: `node-sass` is fine, we use it only in the tests.

mgechev: Would you also add a test case with different quotes and whitespace?

alisd23: Ok, added a couple more. Should cover all the cases in the Regex.

PR #485, opened 2018-01-09
fix(i18n): do not warn with interpolation in i18n element

### LINKED ISSUE #442: i18n check-text edge cases
Some cases where the i18n rule with `check-text` errors, but shouldn't (imo):

```html
<mat-checkbox i18n="@@terms">
  I agree to the
  <a href="#">terms of service</a>
</mat-checkbox>
```

```html
<ng-container i18n="@@minlength">
  Use at least {{ minLength }} characters
</ng-container>
```

mgechev:
Fix #442


mgechev:
Before we merge, let's discuss in #442 if this makes sense.

PR #490, opened 2018-01-15
feat: add `no-input-prefix` rule

alan-agius4:
Not sure, if this might make sense for you guys. But since, I was gonna do it for one of our internal libraries, I though to do it here first. 

Maybe you like it, or maybe not :)

If you'd like it I can also extend it and has it to include other `boolean` prefixes to be excluded from inputs, such as `has`, `should` `can` etc.. And it can be renamed too `no-input-boolean-prefixes`, maybe?

The whole point for the rules is to disable boolean prefixes from input properties. I personally find them a bit dirty, even though they are not listing anyways if one should do them or not in the Angular style guide. But, considering even the native html attributes like `disabled`, `checked` these are never prefixed.

Let me know what are your thoughts anyways :) I'd appreciate any type of feedback Thanks.

### DIFF
--- a/src/noInputPrefixRule.ts
+++ b/src/noInputPrefixRule.ts
@@ -1,0 +1,55 @@
+import * as Lint from 'tslint';
+import * as ts from 'typescript';
+import { sprintf } from 'sprintf-js';
+import { NgWalker } from './angular/ngWalker';
+
+export class Rule extends Lint.Rules.AbstractRule {
+  public static metadata: Lint.IRuleMetadata = {
+    ruleName: 'no-input-prefix',
+    type: 'maintainability',
+    description: 'Input names should not be prefixed with the configured disallowed prefixes.',
+    rationale: `HTML attributes are not prefixed. It's considered best not to prefix Inputs.
+    * Example: 'enabled' is prefered over 'isEnabled'.
+    `,
+    options: {
+      'type': 'array',
+      'items': [
+        { 'type': 'string' }
+      ],
+    },
+    optionExamples: [
+      '["is", "can", "should"]'
+    ],
+    optionsDescription: 'Options accept a string array of disallowed input prefixes.',
+    typescriptOnly: true
+  };
+
+  static FAILURE_STRING: string = 'In the class "%s", the input property "%s" should not be prefixed with %s';
+
+  public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+    return this.applyWithWalker(new InputWalker(sourceFile, this.getOptions()));
+  }
+}
+
+class InputWalker extends NgWalker {
+  visitNgInput(property: ts.PropertyDeclaration, input: ts.Decorator, args: string[]) {
+    const className = (<any>property).parent.name.text;
+    const memberName = (<any>property.name).text as string;
+    const options = this.getOptions() as string[];
+    let prefixLength: number;
+
+    if (memberName) {
+      const foundInvalid = options.find(x => memberName.startsWith(x));
+      prefixLength = foundInvalid ? foundInvalid.length : 0;
+    }
+
+    if (
+      prefixLength > 0 &&
+      !(memberName.length >= prefixLength + 1 && memberName[prefixLength] !== memberName[prefixLength].toUpperCase())
+    ) {
+      const failureConfig: string[] = [Rule.FAILURE_STRING, className, memberName, options.join(', ')];
+      const errorMessage = sprintf.apply(null, failureConfig);
+      this.addFailure(this.createFailure(property.getStart(), property.getWidth(), errorMessage));
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


alan-agius4:
@mgechev any feedback pls?

mgechev:
Looks like you already have a bunch of prefixes in mind. Maybe a generic rule called `no-input-prefix`, which as argument accepts a list of prefixes would be a better idea.

alan-agius4:
Ill try to update my PR in the comings days.

alan-agius4:
PR updated

src/noInputPrefixRule.ts
```diff
@@ -0,0 +1,46 @@
...
+    type: 'maintainability',
+    description: 'Name inputs without the prefix with the provided prefixes',
+    rationale: `HTML attributes are not prefixed. It's condered best not to use 'is', 'should', 'must' prefixes for boolean values inputs.`,
+    options: {
+      'type': 'array',
+      'items': [
+        { 'type': 'string' }
+      ],
+    },
+    optionsDescription: 'Not configurable.',
```
wKoza: `optionsDescription`  requires a description.

src/noInputPrefixRule.ts
```diff
@@ -0,0 +1,46 @@
...
+    typescriptOnly: true
+  };
```
wKoza: Can you add an example with `optionExamples` ?

src/noInputPrefixRule.ts
```diff
@@ -0,0 +1,46 @@
...
+import * as ts from 'typescript';
+import { sprintf } from 'sprintf-js';
+import { NgWalker } from './angular/ngWalker';
+
+export class Rule extends Lint.Rules.AbstractRule {
+  public static metadata: Lint.IRuleMetadata = {
+    ruleName: 'no-input-prefix',
+    type: 'maintainability',
+    description: 'Name inputs without the prefix with the provided prefixes',
+    rationale: `HTML attributes are not prefixed. It's condered best not to use 'is', 'should', 'must' prefixes for boolean values inputs.`,
```
wKoza: condered  => considered ?

src/noInputPrefixRule.ts
```diff
@@ -0,0 +1,46 @@
+import * as Lint from 'tslint';
+import * as ts from 'typescript';
+import { sprintf } from 'sprintf-js';
+import { NgWalker } from './angular/ngWalker';
+
+export class Rule extends Lint.Rules.AbstractRule {
+  public static metadata: Lint.IRuleMetadata = {
+    ruleName: 'no-input-prefix',
+    type: 'maintainability',
+    description: 'Name inputs without the prefix with the provided prefixes',
```
wKoza: can you propose something else ?

src/noInputPrefixRule.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+class InputWalker extends NgWalker {
+  visitNgInput(property: ts.PropertyDeclaration, input: ts.Decorator, args: string[]) {
+    const className = (<any>property).parent.name.text;
+    const memberName = (<any>property.name).text as string;
+    const options = this.getOptions() as string[];
+    let prefixLength = 0;
+
+    if (
```
wKoza: This rule isn't only for boolean Input ?

test/noInputPrefixRule.spec.ts
```diff
@@ -0,0 +1,85 @@
+import { assertSuccess, assertAnnotated } from './testHelper';
+
+describe('no-input-prefix', () => {
+  describe('invalid directive input property', () => {
+    it('should fail, when a component input property is named with is prefix', () => {
+      const source = `
+      @Component()
+      class ButtonComponent {
+        @Input() isDisabled: boolean;
```
wKoza: Can you add a test with a type `string`.

src/noInputPrefixRule.ts
alan-agius4: In reality it can work for all type of inputs

src/noInputPrefixRule.ts
alan-agius4: Inputs names should not be prefixed with the configured disallowed prefixes.

What do you thunk about this?

src/noInputPrefixRule.ts
alan-agius4: Unless we want to limit it to booleans only

alan-agius4:
I pushed some changes following the PR review. The only pending are the 2 last comment.

Should we make this property explicitly for `boolean` type inputs?

wKoza:
Not necessary but in this case, we could change our rational

> HTML attributes are not prefixed. It's considered best not to use 'is', 'should', 'must' prefixes for **boolean** values Inputs.

BTW, you could change this sentence in any case because prefixes are now configurable.

alan-agius4:
I updated the `rational` check if makes sense :)

Thanks for the review!

Issue #491, opened 2018-01-16
Allow disabling of rules in templates and styles


alan-agius4:
At the moment if I add a rule example `trackBy-function` which validates html content there is no way to disable a rule for a specific file or line such as https://palantir.github.io/tslint/usage/rule-flags/

It would be nice to have this feature, in order to disable a rule where one doesn't want to emit a lint error.

mgechev:
This could be caused by the rule naming. Did you try using both `/* tslint disable:trackBy-function */` and the lower case alternative?

alan-agius4:
I can try again, but when I checked tslint was only looking for these in TS
files and not HTML files, causes it uses TypeScript to find comments and it
cannot read comments in Html files.

But surly I can try the lower case later and give you an update.

mgechev:
@alan-agius4 yes, you should add this marker in the TypeScript file referencing your HTML template.

alan-agius4:
That works fine thanks :) I'll happily do a PR. Maybe you can indicate in
which readme it should go?

mgechev:
In the root of the project should be fine. Thanks!

rgant:
It would be great if we could actually disable this rule for specific lines in the template file. I have a situation where some of my ngForOfs are just looping over strings and don't need template-use-track-by-function.

antch:
> It would be great if we could actually disable this rule for specific lines in the template file. I have a situation where some of my ngForOfs are just looping over strings and don't need template-use-track-by-function.

Absolutely. I'm finding that I have to disable the rule altogether as around half of the usages in the codebase use strings/numbers, and if we have to disable the rule for an entire template there's really no point in trying to enforce it.

PR #492, opened 2018-01-16
docs(README): add disable a rule that validates template or styles

alan-agius4:
@mgechev here you go :)

relates to #491

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,6 +256,23 @@ Below you can find a recommended configuration which is based on the [Angular St
 }
 ```
 
+## Disable a rule that validates Template or Styles
+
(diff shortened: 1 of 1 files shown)


rgant:
Did this get removed from the README? 

In any event it seems to work very strangely. If I add the comment `/* tslint:disable:template-use-track-by-function */` to line 12 of the component.ts file, then it disables the check for lines 12+ in the component.html file, but not for lines 1-11.

So I don't think this is a great solution. Any way to add this flag to specific ngForOfs within the template itself?

stephanie-dm:
I agree with @rgant : it's strange behaviour... I didn't realise why my disable rule didn't work, and it was because I didn't put it as the FIRST rule in my *.ts file

Issue #487, opened 2018-01-10
Incorrect formatting with check-interpolation


mgechev:
@dereklin:

> When I run npm run lint -- --fix
> This
> ```
> <h1>
>   {{name}}
> </h1>
> ```
> is turned into this:
> ```
> <h1>
>  {{ {nam }}}
> </h1>
> ```
> Here is my repo: https://github.com/dereklin/nx-demo-jest/tree/try-tslint-config
> The template is at: https://github.com/dereklin/nx-demo-jest/blob/try-tslint-config/apps/app1/src/app/feature1/feature1.component.html

https://github.com/mgechev/tslint-angular/issues/1

wKoza:
Duplicate #452 ?

mgechev:
Here we have a broken auto fix.

mgechev:
I cannot reproduce this issue https://github.com/mgechev/codelyzer/pull/494

wKoza:
What's your OS @dereklin ?

wKoza:
I took a closer look at this problem.  It's the same sort of problem that #452 . It may be a problem of carriage return. I've never been able to catch it with a TU.

mgechev:
@wKoza I thought it might be OS related but we run tests in Windows as well.

dereklin:
@wKoza It's Windows 10.  I encountered the problem when I was trying out tslint-angular.

wKoza:
I will test with my PC at work. But, if I remember correctly, This problem is not catchable with a TU currently .

wKoza:
I tested  with Windows. I  reproduce the problem with Webstrom. 

![img1](https://image.ibb.co/c5z8sR/Capture1.png)

![img2](https://image.ibb.co/daFsK6/Capture2.png)

When I change the file encoding to LF instead of CRLF, it's work:

![img3](https://image.ibb.co/bs53Rm/Capture3.png)

I'll search a fix ...

mgechev:
Thanks for looking into it!

wKoza:
Correct this issue isn't easy. In this case, the problem happens when the corrected string  is injected into the file with Tslint. There is a lag between `\n` et `\r\n`.

mgechev:
Will move this issue to a future release.

agrinko:
I can add another case to this issue.
Angular-whitespace autofix breaks my component in the following piece of template:
```
  <span *ngSwitchCase="'interval'">
[...]

Issue #452, opened 2017-11-04
"angular-whitespace" rule false-positive (when using string with pipe, ngx-translate)


Ritzlgrmft:
My code is par example:

```html
<button>
  {{ "button.label" | translate }}
</button>
```

This is complained by codelyzer: `The pipe operator should be surrounded by one space on each side, i.e. " | ".`
However, if I add a `span`, codelyzer is satisfied: 

```html
<button>
  <span>{{ "button.label" | translate }}</span>
</button>
```

The same problem was already reported in #415. It was fixed in release 3.2.2, but in 4.0.1, it happens again.

FabienDehopre:
I have this issue even when using version 3.2.2

FabienDehopre:
BTW, the error happens randomly with 3.2.2. 
I made a change (removed an empty line) and the error did not happen.

FabienDehopre:
Any news about this issue?
It still happens on the latest release (4.0.2)

mgechev:
Here are the changes from [4.0.2](https://github.com/mgechev/codelyzer/blob/master/CHANGELOG.md#402).

FabienDehopre:
OK. I read the changelog for the 4.0.2. 
My question was when would this issue (related to the check-pipe) be solved?

jeneg:
Got the same issue on 4.0.2. @mgechev Is it going to be fixed?

mgechev:
Let's drop the `check-pipe` as @wKoza suggested https://github.com/mgechev/codelyzer/issues/460.

slimlime:
I had an issue with angular-whitespace or one of the fixers that seemed to identify angular moustache templates in my typescript files (no templates) and 

```ts
import {
  FormControl,
  FormGroup,
  FormBuilder,
  Validators,
  FormArray
} from "@angular/forms";

```
became 
```ts
imp; }}t{{ { {  }}{  }} {  }} {  }}
  }}ormArray,
  FormBuilder,
  FormGroup,
  Validators;
} from; '@angular/forms';
```

I had ridiculously run the tslint --fix autofix while working out how to use tslint and ended up replacing even more `}}` 🤣.

I'm still trying to track things down.

PR #495, opened 2018-01-18
add tests in relation with #452 and #487

wKoza:
related to  #452 and #487


mgechev:
@wKoza yes, the tests are failing on Windows.

wKoza:
Yes @mgechev , and we can also reproduce on Linux for CRLF

mgechev:
Let's close for now.

PR #498, opened 2018-01-23
New rule: disallow full import of RxJS operators

rokerkony:
The reason behind this PR is better described in my article:
https://medium.com/p/prevent-coding-mistakes-write-a-tslint-rule-specific-for-your-own-project-987d26f91647

Shortly: it can decrease a size of a bundle when RxJS operators are imported in a wrong way

_I know it is not exactly Angular related, but Angular requires RxJS as a dependency. Therefore I think it could be part of the Styleguide._

### DIFF
--- a/src/disallowFullImportRxjsOperatorsRule.ts
+++ b/src/disallowFullImportRxjsOperatorsRule.ts
@@ -1,0 +1,75 @@
+import { ImportDeclaration, SourceFile } from 'typescript';
+import { IRuleMetadata, Replacement, RuleFailure, Rules, RuleWalker } from 'tslint';
+
+const disallowedImports: string[] = [
+  // lettable import
+  '\'rxjs/operators\'',
+  '"rxjs/operators"',
+  // not-lettable import
+  'rxjs/add/operator',
+];
+
+export class Rule extends Rules.AbstractRule {
+  public static metadata: IRuleMetadata = {
+    ruleName: 'disallow-full-import-rxjs-operators',
+    type: 'maintainability',
+    description: 'RxJS operators should be imported from it\'s own scope.',
+    rationale: 'Helps to improve an angular bundle size. Angular has a dependency' +
+    ' on RxJS and it helps to not increase a final bundle size.',
+    options: null,
+    optionsDescription: 'Not configurable.',
+    typescriptOnly: true,
+  };
+
+  public static FAILURE_STRING = 'RxJS operators should be imported from their own scope.';
+
+  public apply(sourceFile: SourceFile): RuleFailure[] {
+    return this.applyWithWalker(new Walker(sourceFile, this.getOptions()));
+  }
+}
+
+class Walker extends RuleWalker {
+  public visitImportDeclaration(importDeclaration: ImportDeclaration): void {
+    const importDeclarationText: string = importDeclaration.getText();
+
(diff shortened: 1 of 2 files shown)


intellix:
I was under the impression this was ok: `import {delay, tap} from 'rxjs/operators`
Also, can't you already achieve this with `import-blacklist` from tslint?

rokerkony:
@intellix Let me try this once more in the evening but when I checked on the codebase of https://github.com/erento with `source-map-explorer` the outcome was different.

The advantage is that it has a fixer as well, so it can automatically fix the imports.

intellix:
```ts
ng new rxjs-test && cd rxjs-test
npm i source-map-explorer
ng build -prod -sm -oh=false
./node_modules/.bin/source-map-explorer dist/main.bundle.js
```

**app.component.ts:**
```ts
import { Component } from '@angular/core';
import { timer } from 'rxjs/observable/timer';
import { tap } from 'rxjs/operators';

@Component({
  selector: 'app-root',
  templateUrl: './app.component.html',
  styleUrls: ['./app.component.css']
})
export class AppComponent {
  title = 'app';
  timer = timer(100, 100).pipe(tap(console.log));
}
```

**Before using above (20.67kb):**
screenshot 2018-01-23 14 48 50


**After using above (27.01kb):**
screenshot 2018-01-23 14 52 52

lazarljubenovic:
I don't think this belongs to codelyzer. Check out [`rxjs-lint-rules`](https://www.npmjs.com/package/rxjs-tslint-rules).

rokerkony:
@intellix I don't know what you mean with before and after... but from curiosity I run the same example you posted and these are my results:
`import { tap } from 'rxjs/operators/tap';` -> **24.49 KB**
`import { tap } from 'rxjs/operators';` -> **27.01 KB**

So I have a bit different number then you in the first line — it shows it is better to import directly from `rxjs/operators/tap`.

@lazarljubenovic I still think it would be valuable for other angular users but I'm not pushing for it hard, so when you agree I should close it I will :)

intellix:
ok so it seems more stuff is pulled in. I was thinking it was more a case of importing **all** vs **a single** operator

wKoza:
I'm not sure this rule is durable. There are two things:
- Rxjs explains that the best way to use lettable / pipeable is:

```javascript
import { map, filter, mergeMap, tap } from 'rxjs/operators';
```

see : https://github.com/ReactiveX/rxjs/issues/3018#issuecomment-347255928

- Our problem seems to be related to Webpack 3. Webpack 4 will improve it.

see: https://github.com/angular/angular-cli/issues/9069#issuecomment-355353286

So, using the deep import is a workaround. I'm not sure that a provisional rule has its place in codelyzer.

lucasvanhalst:
They will remove the deep imports in rxjs 6, so don't think this rule should become a thing.

See here: https://github.com/ReactiveX/rxjs/blob/master/CHANGELOG.md#600-alpha1-2018-01-12

> operators: Pipeable operators must now be imported from rxjs like so: import { map, filter, switchMap } from 'rxjs/operators';. No deep imports.

rokerkony:
Thank you @wKoza and @lucasvanhalst for great arguments, I will close this PR as it really doesn't make sense when RxJS@6 will not have this problem anymore...

PR #506, opened 2018-01-31
WIP fix(rule): add i18n test with pural

### LINKED ISSUE #504: i18n rule pluralization does not work
**Dependencies**
* codelyzer: 4.1.0
* tslint: 5.9.1
* angular: 5.0.0
* typescript: 2.4.2

**Affected rule**
i18n  
`"i18n": [true, "check-text"]`

**Problem**
The i18n rule does not recognize an i18n tag with pluralization rules correctly.
The problem appears with and without a meaning|descripion.

**Code to reproduce problem**
`<span i18n>Updated {minutes, plural, =0 {just now} =1 {one minute ago} other {{{minutes}} minutes ago}}</span>`
This is the official example from the angular i18n documentation.
https://angular.io/guide/i18n#translate-singular-and-plural

**Error message**
Each element containing text node should have an i18n attribute

wKoza:
Closes #504

### DIFF
--- a/test/i18nRule.spec.ts
+++ b/test/i18nRule.spec.ts
@@ -131,6 +131,18 @@ describe('i18n', () => {
       assertSuccess('i18n', source, ['check-text']);
     });
 
+    it.only('should work with plural', () => {
+      const source = `
+      @Component({
+        template: \`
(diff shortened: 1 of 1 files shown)


wKoza:
I'm not sure of my fix. Seems to be doing the job.

PR #509, opened 2018-02-03
feat(rule): add templateConditionalComplexityRule

### LINKED ISSUE #508: [Feature Request] Limiting template condition complexity
We've got a payment process with conditions to reveal different steps of said process to push the user through it. An example of an *ngIf condition for revealing the Card entry step to the user: 

```html
<div *ngIf="step === 'amount' || (step === 'card' && transaction.action !== 'iframe-sent' && transaction.action !== 'html-provided' && transaction.action !== 'redirect-to-url')">
  Enter your card details
</div>
```

This amount of condition complexity is dangerous because it's breakage can only be detected via E2E or Manual QA. During refactor another developer can completely break payments.

I think one way Codelyzer can help here is allowing us to limit the complexity of an `*ngIf` condition allowed in the view. The above example would throw that it's too complex and the condition needs to be moved into the Component class, of which can be unit tested:

```ts
get canShowCardStep() {
  return this.step === 'amount' || (this.step === 'card' && this.transaction.action !== 'iframe-sent' && this.transaction.action !== 'html-provided' && this.transaction.action !== 'redirect-to-url;
}
```

```html
<div *ngIf="canShowCardStep">
  Enter your card details
</div>
```

wKoza:
closes #508

### DIFF
--- a/src/complexityRule.ts
+++ b/src/complexityRule.ts
@@ -1,0 +1,560 @@
+// based on https://github.com/wimpyprogrammer/conditional-expression-parser - LICENSE MIT - Copyright (c) 2015 Andrew Keller
+import * as Lint from 'tslint';
+import * as ts from 'typescript';
+import * as ast from '@angular/compiler';
+import { sprintf } from 'sprintf-js';
+import { BasicTemplateAstVisitor } from './angular/templates/basicTemplateAstVisitor';
+import { NgWalker } from './angular/ngWalker';
+import { isObject } from 'util';
+
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: 'complexity',
+        type: 'functionality',
+        description: 'The condition complexity shouldn\'t exceed a rational limit in a template.',
+        rationale: 'An important complexity complicates the tests and the maintenance.',
+        options: {
+            type: 'array',
+            items: {
+              type: 'string'
+            },
+            minLength: 0,
+            maxLength: 2,
+          },
+          optionExamples: [
+            'true',
+            '[true, 4]'
+          ],
+        optionsDescription: 'Not configurable.',
+        typescriptOnly: true,
+        hasFix: false
+
+    };
+
+    // tslint:disable-next-line:max-line-length
+    static COMPLEXITY_FAILURE_STRING = 'The condition complexity (cost \'%s\') exceeded the defined limit (cost \'%s\'). The conditional expression should be move in the component\'s template.';
+
+    static COMPLEXITY_MAX = 3;
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+
+        return this.applyWithWalker(
+            new NgWalker(sourceFile,
+                this.getOptions(), {
+                    templateVisitorCtrl: ComplexityTemplateVisitor,
+                }));
+    }
+}
+
+
+class ComplexityTemplateVisitor extends BasicTemplateAstVisitor {
+
+    visitDirectiveProperty(prop: ast.BoundDirectivePropertyAst, context: BasicTemplateAstVisitor): any {
+
+        if (prop.sourceSpan) {
+            const directive = (<any>prop.sourceSpan).toString();
+
+            if (directive.startsWith('*ngIf')) {
+
+                const expr = directive.split(/\*ngIf\s*=\s*/)[1].replace(/[\n\r]/g, ''); // extract expression and drop character new line
+                const expression = new Expression(expr);
+
+                let complexity = 0;
+                let currentExpression = expression;
+
+                for (let i = 0; i < currentExpression.conditions.length; i++) {
+                    if (currentExpression.conditions[i] instanceof Condition) {
+                        complexity++;
+                    } else if (currentExpression.conditions[i] instanceof Expression) {
+                        currentExpression = (currentExpression.conditions[i]);
+                        i = 0;
+                    }
+                }
+
+                const options = this.getOptions();
+                const complexityMax: number = options.length ? options[0] : Rule.COMPLEXITY_MAX;
+
+                if (complexity > complexityMax) {
+                    const span = prop.sourceSpan;
+                    let failureConfig: string[] = [String(complexity), String(Rule.COMPLEXITY_MAX)];
+                    failureConfig.unshift(Rule.COMPLEXITY_FAILURE_STRING);
+                    this.addFailure(this.createFailure(span.start.offset, span.end.offset - span.start.offset,
+                        sprintf.apply(this, failureConfig))
+                    );
+                }
+            }
+        }
+        super.visitDirectiveProperty(prop, context);
+    }
+}
+
+
+class Utils {
+
+    static tokensLineBreak = ['\\r', '\\n', '\\r\\n'];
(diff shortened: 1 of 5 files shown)


src/complexityRule.ts
```diff
@@ -0,0 +1,647 @@
...
+    };
+}
+
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: 'complexity',
+        type: 'functionality',
+        description: 'Ensure that the two-way data binding syntax is correct.',
+        rationale: 'The parens "()" should have been inside the brackets "[]".',
```
intellix: I guess both description and rationale are a WIP? :)

wKoza: Yes, a very early WIP ;)

src/complexityRule.ts
```diff
@@ -43,8 +58,7 @@ class ComplexityTemplateVisitor extends BasicTemplateAstVisitor {
 
             if (directive.startsWith('*ngIf')) {
 
-                const directiveMatch = directive.match(ngIfExpressionRe);
-                const expr = directiveMatch && directiveMatch[1];
+                const expr = directive.split(/\*ngIf\s*=\s*/)[1].replace(/[\n\r]/g, ''); // extract expression and drop caractere new line
```
lazarljubenovic: caractere -> character

wKoza: done

mgechev:
I will take a detailed look tomorrow. Would suggest to rename the rule to `template-conditional-complexity`. In future we may implement a rule which computes a complexity value for the entire template, considering all the operations inside, similarly to cyclomatic complexity that I mentioned previously.

wKoza:
I'm agree with you. It's just the first step before the determination of O.

mgechev:
I think it makes sense to have both. The cyclomatic complexity is more restrictive, this will be perfect for more conservative users.

wKoza:
I found a bug in the analysis of the expression. I'm going to invest before your review. I'll let you know.

wKoza:
@mgechev , it's ok for a review.

mgechev:
@wKoza awesome job! Would you switch to 2 spaces instead of 4 to be consistent with most of the other rules?

src/templateConditionalComplexityRule.ts
```diff
@@ -0,0 +1,562 @@
...
+        type: 'string'
+      },
+      minLength: 0,
+      maxLength: 2,
+    },
+    optionExamples: [
+      'true',
+      '[true, 4]'
+    ],
+    optionsDescription: 'Not configurable.',
```
mgechev: It's configurable.

src/templateConditionalComplexityRule.ts
```diff
@@ -0,0 +1,562 @@
...
+    typescriptOnly: true,
+    hasFix: false
+
```
mgechev: ✂️

src/templateConditionalComplexityRule.ts
```diff
@@ -0,0 +1,562 @@
...
+enum COND_TYPE {
+  TYPE_AND = 'AND',
+  TYPE_OR = 'OR',
+  TYPE_XOR = 'XOR'
+}
+
+class Operator {
+
+  private type: COND_TYPE;
+
```
mgechev: ✂️

src/templateConditionalComplexityRule.ts
```diff
@@ -0,0 +1,562 @@
...
+    conditions.forEach((c) => {
+      evalPaths[0].push({condition: c, result: result});
+    });
+    return evalPaths;
+  }
+
+
+}
+
+class Expression {
```
mgechev: Curious why you didn't choose to use the expression parser of Angular?

src/templateConditionalComplexityRule.ts
```diff
@@ -0,0 +1,562 @@
...
+
+class Utils {
+
+  static tokensLineBreak = ['\\r', '\\n', '\\r\\n'];
+  static tokensAnd = ['&&', '\\sAND\\s']; // "&&" or "AND"
+  static tokensOr = ['\\|\\|', '\\sOR\\s']; // "||" or "OR"
+  static tokensXor = ['\\^', '\\sXOR\\s']; // "^" or "XOR"
+  static tokensAndOrXor = [].concat(Utils.tokensAnd, Utils.tokensOr, Utils.tokensXor);
+
+
```
mgechev: ✂️

src/templateConditionalComplexityRule.ts
```diff
@@ -0,0 +1,562 @@
...
+  }
+
+  isOr() {
+    return this.type === COND_TYPE.TYPE_OR;
+  }
+
+  isXor() {
+    return this.type === COND_TYPE.TYPE_XOR;
+  }
+
```
mgechev: ✂️

test/complexityRule.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+      class Bar {}
+      `;
+            assertSuccess('template-conditional-complexity', source);
+        });
+
+        it('should work with a level of complexity customisable', () => {
+            let source = `
+      @Component({
+        template: \`
+        <div *ngIf="a === '3' || (b === '3' && c.d !== '1' && e.f !== '6' && q !== g)">
```
mgechev: This will fail to parse if the expression is:

```
a === '3' || b === '3' && c.d !== '1' && e.f !== '6' && q !== g
```
right?

test/complexityRule.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+import { Replacement } from 'tslint';
+import { expect } from 'chai';
+
+describe('complexity', () => {
+    describe('success', () => {
+        it('should work with a lower level of complexity', () => {
+            let source = `
+      @Component({
+        template: \`
+        <div *ngIf="a === '1' || (b === '2' && c.d !== e)">
```
mgechev: This will fail to parse if the expression is:

```
a === '1' || b === '2' && c.d !== e
```
right?

mgechev:
@wKoza left some comments. I'm mostly curious why you prefer to not use the Angular's expression parser.

test/complexityRule.spec.ts
wKoza: No. I added a test with this expression.

src/templateConditionalComplexityRule.ts
wKoza: I didn't use the parser for three main reasons:
- I didn't find easily the information I needed. I can't say that this parser is very "user friendly".
- Without the expression parser of Angular, we are bound only on the public API.
- Our analysis is purely static

mgechev: If we use custom parser we can hit a problem because of the DSL Angular supports, which has extra syntax on top of JavaScript.

We already depend on the expression parser in the `NgWalker`. I don't think using it will make our analysis dynamic?

What information is missing there?

wKoza: Hi @mgechev,
Ok, I've found my holy graal ! thx.

mgechev: Beautiful! Will take a final look tomorrow and merge.

PR #511, opened 2018-02-08
fix: cannot find rules

alan-agius4:
At the moment doing something like
```json
{
  "extends": ["codelyzer"],
  "rules": {
    "angular-whitespace": [true, "check-interpolation", "check-semicolon"],
    "banana-in-box": true,
    "templates-no-negated-async": true,
    "directive-selector": [true, "attribute", "sg", "camelCase"],
  }
}
```

Will not work and you'll end up with loads of errors example
```bash
Could not find implementations for the following rules specified in the configuration:
    angular-whitespace
    banana-in-box
    contextual-life-cycle
```

This is because tslint requires a `rulesDirectory` to resolve the rules.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -28,3 +28,6 @@ export { Rule as UseViewEncapsulationRule } from './useViewEncapsulationRule';
 export { Rule as TemplatesNoNegatedAsync } from './templatesNoNegatedAsyncRule';
 export { Rule as TrackByFunctionRule } from './trackByFunctionRule';
 export * from './angular/config';
+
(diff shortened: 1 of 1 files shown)


mgechev:
Would you share the source where you found that exporting `rulesDirectory` is required?

alan-agius4:
Actually i found it in some source code of other lint rules libraries
example;

https://github.com/buzinas/tslint-eslint-rules/blob/master/index.js

https://github.com/ajafff/tslint-consistent-codestyle/blob/master/rules/index.ts

PR #512, opened 2018-02-10
feat(rule): cyclomatic complexity rule

wKoza:
related to #508

### DIFF
--- a/test/templateCyclomaticComplexityRule.spec.ts
+++ b/test/templateCyclomaticComplexityRule.spec.ts
@@ -1,0 +1,84 @@
+// tslint:disable:max-line-length
+import { assertSuccess, assertAnnotated } from './testHelper';
+import { Replacement } from 'tslint';
+import { expect } from 'chai';
+
+describe('cyclomatic complexity', () => {
+    describe('success', () => {
+        it('should work with a lower level of complexity', () => {
+            let source = `
+      @Component({
+        template: \`
+        <div *ngIf="a === '1'">
+            <li *ngFor="let person of persons; trackBy: trackByFn">
+              {{ person.name }}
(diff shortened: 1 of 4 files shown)


test/templateCyclomaticComplexityRule.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+      @Component({
+        template: \`
+        <div *ngIf="a === '1'">
+            <li *ngFor="let person of persons; trackBy: trackByFn">
+             <div *ngIf="a === '1'">{{ person.name }}</div>
+                <div [ngSwitch]="person.emotion">
+                    <app-happy-hero    *ngSwitchCase="'happy'"    [hero]="currentHero"></app-happy-hero>
+                    <app-sad-hero      *ngSwitchCase="'sad'"      [hero]="currentHero"></app-sad-hero>
+                    <app-confused-hero *ngSwitchCase="'confused'" [hero]="currentHero"></app-confused-hero>
+                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~
```
intellix: Isn't this test exactly the same as above?

wKoza: In the previous test, We have set a complexity of 7.

wKoza:
@mgechev , it's a mix of [gocyclo](https://github.com/fzipp/gocyclo/blob/master/gocyclo.go) and [checkstyle](https://github.com/maddingo/checkstyle/blob/master/src/checkstyle/com/puppycrawl/tools/checkstyle/checks/metrics/CyclomaticComplexityCheck.java). 
Cyclomatic complexity is computed by adding the nodes (directives). I don't add the complexity of Boolean Expressions (LAND , LOR).

PR #513, opened 2018-02-11
docs(README): add npm version badge

alan-agius4:
This is quite helpful as in order to check what’s the latest version one doesn’t need to check the chane log or package.json


wKoza:
LGTM!

PR #538, opened 2018-03-08
fix(537): no-output-named-after-standard-event checks output rename

### LINKED ISSUE #537: `no-output-named-after-standard-event` Does Not Check Output Rename
```TypeScript
@Output('click') onVoted = new EventEmitter<boolean>();
```

Is not caught by the `no-output-named-after-standard-event` rule.

gbilodeau:
Fixes #537


mgechev:
Thanks for the fix!

Issue #443, opened 2017-10-24
Nested i18n linting


intellix:
A couple of us within our team accidentally nest i18n properties which will cause xi18n extraction to fail: 
```
<div i18n>
  <span i18n>First name</span>
</div>
```

The error thrown during xi18n is:
```
An error occured during the i18n extraction:
Error: Could not mark an element as translatable inside a translatable section
```

Not sure if this should fail Angular compilation or if there should be a linting rule to help identify these before they get merged. Thoughts?

mgechev:
@intellix at first, to me it looks like it's the responsibility of xi18n to throw an error.

Issue #530, opened 2018-03-04
i18n rule with check-id and check-text does not handle the ng-container tag


sgallen:
Reference: PR #506

Today I updated to version 4.2.0 of Codelyzer, as I was finding the linter was incorrectly reporting errors with our use of `select` ICU expressions (see: https://angular.io/guide/i18n#translate-select).  PR #506 seems to have addressed those issues (thanks!).  Unfortunately, I'm now finding that when I use `ng-container` tags with i18n ids I get linting errors, that I don't expect (see: https://angular.io/guide/i18n#translate-text-without-creating-an-element).

I've cloned the repo, locally adding 2 tests to the suite.  They're identical tests except for the use of different tags (`ng-container` vs `span`). As you'll see further below `ng-container` fails while `span` works fine:
```
 66     it('NEW [ng-container]: should work with proper id', () => {
 67       const source = `
 68       @Component({
 69         template: \`
 70           <ng-container i18n="@@foo">Text</ng-container>
 71         \`
 72       })
 73       class Bar {}
 74       `;
 75       assertSuccess('i18n', source, ['check-id', 'check-text']);
 76     });
 77
 78     it('NEW [span]: should work with proper id', () => {
 79       const source = `
 80       @Component({
 81         template: \`
 82           <span i18n="@@foo">Text</span>
 83         \`
 84       })
 85       class Bar {}
 86       `;
 87       assertSuccess('i18n', source, ['check-id', 'check-text']);
 88     });
```
This is the output:
```
node@3bb72aa7d7e3:/codelyzer$ npm run test
npm info it worked if it ends with ok
npm info using npm@5.3.0
npm info using node@v8.5.0
npm info lifecycle codelyzer@4.2.1~pretest: codelyzer@4.2.1
npm info lifecycle codelyzer@4.2.1~test: codelyzer@4.2.1

> codelyzer@4.2.1 test /codelyzer
> rimraf dist && tsc && cp -r test/fixtures dist/test && mocha dist/test --recursive



  i18n
    check-id
      ✓ should work with proper id (50ms)
      ✓ should work with proper i18n attribute
      ✓ should work with proper id
      ✓ should work with proper id
      1) NEW [ng-container]: should work with proper id
      ✓ NEW [span]: should work with proper id
      ✓ should fail with missing id string
      ✓ should fail with missing id
      ✓ should fail with missing id


  8 passing (130ms)
  1 failing

  1) i18n check-id NEW [ng-container]: should work with proper id:
[...]

PR #546, opened 2018-03-22
Add support for nested i18n items and removed ng-container restriction

rwlogel:
This fix will address #530 regarding `ng-container` checking.  I verified that the unit test added for #506 still passes.

It will also address another problem that my team was having which was causing the `check-text` to report failure for nested items like this:

```
<ul i18n="@@myList">
    <li>List Item 1</li>
    <li>List Item 2</li>
    <li>List Item 3</li>
</ul>
```
This is valid syntax resulting in the following xlf output:

```
<trans-unit id="myList" datatype="html">
    <source>
        <x id="START_LIST_ITEM" ctype="x-li" equiv-text="&lt;li&gt;"/>List Item 1<x id="CLOSE_LIST_ITEM" ctype="x-li" equiv-text="&lt;/li&gt;"/>
        <x id="START_LIST_ITEM" ctype="x-li" equiv-text="&lt;li&gt;"/>List Item 2<x id="CLOSE_LIST_ITEM" ctype="x-li" equiv-text="&lt;/li&gt;"/>
        <x id="START_LIST_ITEM" ctype="x-li" equiv-text="&lt;li&gt;"/>List Item 3<x id="CLOSE_LIST_ITEM" ctype="x-li" equiv-text="&lt;/li&gt;"/>
    </source>
</trans-unit>
```
Since they were touching the same area of the code I just put the changes together, but if necessary I can submit as multiple requests.

### DIFF
--- a/src/i18nRule.ts
+++ b/src/i18nRule.ts
@@ -73,13 +73,12 @@ class I18NTextVisitor extends BasicTemplateAstVisitor implements ConfigurableVis
   }
 
   visitElement(element: ast.ElementAst, context: BasicTemplateAstVisitor) {
-    if (element.name !== 'ng-container') {
-      this.hasI18n = element.attrs.some(e => e.name === 'i18n');
-    }
+    const originalI18n = this.hasI18n;
+    this.hasI18n = originalI18n || element.attrs.some(e => e.name === 'i18n');
     this.nestedElements.push(element.name);
     super.visitElement(element, context);
     this.nestedElements.pop();
-    this.hasI18n = false;
+    this.hasI18n = originalI18n;
   }
 
   getOption(): Option {
--- a/test/i18nRule.spec.ts
+++ b/test/i18nRule.spec.ts
@@ -24,6 +24,18 @@ describe('i18n', () => {
       `;
       assertSuccess('i18n', source, ['check-id']);
     });
+    
+    it('should work with proper id on ng-container', () => {
+      const source = `
+      @Component({
+        template: \`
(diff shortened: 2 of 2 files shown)


rwlogel:
One other thing to note regarding nested items, elements can't have `i18n` tags when they are contained within another element with `i18n` tags.  So this unit test is allowing an invalid syntax to pass:

```
182     it('should work with multiple valid elements', () => {
183       const source = `
184       @Component({
185         template: \`
186           <div>{{text}}</div>
187           <div i18n>
188             Text
189             <span i18n>foo</span>
190           </div>
191         \`
192       })
193       class Bar {}
194       `;
195       assertSuccess('i18n', source, ['check-text']);
196     });
```
If this component was extracted using `ng xi18n` this would be the result:

```
ERROR in : Error: Could not mark an element as translatable inside a translatable section ("
    <div i18n>
        Text
        [ERROR ->]<span i18n>foo</span>
    </div>
"
```

The `check-text` directive could easily be updated to alert for this or a new directive could be created to alarm for this instead.  I would be willing to add either solution.

intellix:
Great, should fix this which was closed https://github.com/mgechev/codelyzer/issues/443

rwlogel:
Just to clarify this change won't fix #443 but adding support for that would be pretty straight forward.

rwlogel:
@wKoza Do you know when the next release will be?  My team was hoping to use this fix in our project.

wKoza:
I hope we can make the release for the end of week.

rwlogel:
@wKoza that would be great.  Thank you.

PR #531, opened 2018-03-04
chore: drop dependencies

mgechev:
Fix #525

PR #552, opened 2018-03-28
Bump Angular 6

### LINKED ISSUE #525: Drop @angular/platform-browser-dynamic from peerDependencies

wKoza:
prepares the transition to Angular 6 ...

Supersedes #531 
Fix #525


mgechev:
@hansl confirmed that the current release works with Angular version 6. We can still publish these changes once v6 is out.

PR #556, opened 2018-04-04
fix(rule): inline template max lines rule

NagRock:
* Rule now only validates inline template and styles, before this fix it was looking also into template / style files


wKoza:
LGTM

PR #557, opened 2018-04-08
Add rule in order to enforce the declaration of a selector

### LINKED ISSUE #551: [Feature Request] Component Selector Required
Quick feature request... I'd love it if there was a rule I could enable that would force people to always include a selector for any @Component. Currently if it's for a dialog (dynamically created component) or routed page, you don't technically need one, but it makes the DOM a bit messy.

Perhaps it could be added to the existing component-selector rule?

wKoza:
closes #551


mgechev:
I've been thinking about this. Not sure if introducing a boolean flag will be very intuitive to developers. I'd always wonder what this `true` stands for, especially after the other enumerable properties.

I'm thinking that it's worth it to introduce a separate rule. At first, it looks like an overhead, however, the more I think about it, the better the idea sounds to me. It'll be more flexible and obvious.

wKoza:
I'm agree, use a 'magic' boolean is never a good idea. Well, I will change that very soon.

mgechev:
@wKoza looks good to me! Will take a further look tomorrow and merge.

src/enforceComponentSelectorRule.ts
```diff
@@ -0,0 +1,49 @@
...
+    constructor(sourceFile: ts.SourceFile, private rule: Rule) {
+        super(sourceFile, rule.getOptions());
+    }
+
+    visitNgComponent(metadata: ComponentMetadata) {
+        if (!metadata.selector) {
+            let failureConfig: string[] = [metadata.controller.name.text];
+            failureConfig.unshift(Rule.SELECTOR_FAILURE);
+            this.generateFailure(metadata.decorator.getStart(), metadata.decorator.getWidth(), failureConfig);
+        }
```
mgechev: `super.visitNgComponent`

src/enforceComponentSelectorRule.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export class EnforceComponentSelectorValidatorWalker extends NgWalker {
+
+    constructor(sourceFile: ts.SourceFile, private rule: Rule) {
+        super(sourceFile, rule.getOptions());
+    }
+
+    visitNgComponent(metadata: ComponentMetadata) {
+        if (!metadata.selector) {
+            let failureConfig: string[] = [metadata.controller.name.text];
```
mgechev: Could be a constant.

rafaelss95:
Not related to this PR, but what's the standard in repo: 2 or 4 spaces?

mgechev:
2. We should run prettier to format everything properly.

PR #559, opened 2018-04-16
feat(rule): add no-life-cycle-call rule

### LINKED ISSUE #427: Feature request: disallow explicit calls to lifecycle hooks
Disallow code like

```
method() {
  this.ngOnDestroy()
}
ngOnDestroy() { /* cleanup code here */}
```

Instead, we should write something like

```
cleanup() { /* cleanup code here */}
method() {
  this.cleanup()
}
ngOnDestroy() {
  this.cleanup()
}
```

As [Rob put it](https://gitter.im/angular/angular?at=59daa728f7299e8f53c2c4f3), "don't call your own hooks, because one day, you'll be debugging something and you'll be really confused why hooks are firing unexpectedly/out of order/whatever".

Thought it would be a nice addition to codelyzer.

rafaelss95:
Closes #427.


rafaelss95:
@mgechev can you take a look? Thanks.

src/noLifeCycleCallRule.ts
```diff
@@ -0,0 +1,59 @@
...
+import * as ts from 'typescript';
+import { sprintf } from 'sprintf-js';
+import { NgWalker } from './angular/ngWalker';
+
+export class Rule extends Lint.Rules.AbstractRule {
+  public static metadata: Lint.IRuleMetadata = {
+    ruleName: 'no-life-cycle-call',
+    type: 'maintainability',
+    description: 'Disallows explicit calls to lifecycle hooks.',
+    rationale: 'Explicit calls to lifecycle hooks is confusing.',
```
wKoza: We should be more specific in saying that it is Angular's responsibility to invoke these methods.

src/noLifeCycleCallRule.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  static FAILURE_STRING: string = 'Avoid explicitly calls to lifecycle hooks in class "%s"';
+
+  public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+    return this.applyWithWalker(new ExpressionCallMetadataWalker(sourceFile, this.getOptions()));
+  }
+}
+
+export class ExpressionCallMetadataWalker extends NgWalker {
+  private readonly lifecycles: (keyof typeof LifecycleHooks)[] = [
```
mgechev: Let's add them to a `Set<keyof typeof LifecycleHooks>`.

src/noLifeCycleCallRule.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  visitCallExpression(node: ts.CallExpression): void {
+    this.validateCallExpression(node);
+    super.visitCallExpression(node);
+  }
+
+  private validateCallExpression(node: ts.CallExpression): void {
+    const name = (node.expression as any).name;
+
+    if (!name || this.lifecycles.indexOf(name.text.slice(2)) === -1) {
```
mgechev: This may still return some false positives if the method is called `myOnInit` (not that it's a good idea to name a method this way).

We can introduce a local `LifecycleHooksMethods` union type:

```ts
type LifecycleHooksMethods = 'ngOnInit' | ...;
```

and perform the check with `===` instead of `indexOf`.

rafaelss95: Good catch! I've fixed it, please take a look.

src/noLifeCycleCallRule.ts
rafaelss95: Hey @wKoza, I agree, could you suggest a better _rationale_?

rafaelss95:
Something that left me in doubt when I was implementing this rule was how to name this. The name of the rules that contain the word _lifecycle_ treated it as two separate words: `contextual-life-cycle` and` use-life-cycle-interface`.

However, [Angular's docs calls it "Lifecycle"](https://angular.io/guide/lifecycle-hooks#lifecycle-hooks).

So, what's the correct one? :)

src/noLifeCycleCallRule.ts
mgechev: Sure, you can add tests.

mgechev:
@rafaelss95 let's use `lifecycle` like Angular docs do.

src/noLifeCycleCallRule.ts
wKoza: Just 'Explicit calls to lifecycle hooks is confusing. Invoke lifecycle hooks is the responsability of Angular'

mgechev: Let's also change this:

>Explicit calls to lifecycle hooks could be confusing. Invoke lifecycle hooks is the responsability of Angular.

My only update is "is confusing" to "could be confusing".

mgechev:
@wKoza, we're good here?

wKoza:
@rafaelss95 has to modify the rational and it's good

rafaelss95:
Done. @mgechev @wKoza.

mgechev:
Awesome job! I can publish this on Friday. I have a late flight after ngconf.

rafaelss95:
@mgechev Do you pretend to publish this in `4.4.0 - Ken Thompson`?

rafaelss95:
Oops, my bad. I just noticed that it was published, just not mentioned in `CHANGELOG`.

mgechev:
Thanks for the reminder! I will add this tonight.

PR #563, opened 2018-04-19
feat(rule): add no-conflicting-life-cycle-hooks rule

### LINKED ISSUE #560: Prevent to implement OnChanges and DoCheck on the same class
According to [DoCheck#description](https://angular.io/api/core/DoCheck#description):

> Note that a directive typically should not use both DoCheck and OnChanges to respond to changes on the same input, as ngOnChanges will continue to be called when the default change detector detects changes.

I thought it was a good idea to add a rule to prevent this.

rafaelss95:
Closes #560.

### DIFF
--- a/src/noConflictingLifeCycleHooksRule.ts
+++ b/src/noConflictingLifeCycleHooksRule.ts
@@ -1,0 +1,72 @@
+import { sprintf } from 'sprintf-js';
+import * as Lint from 'tslint';
+import * as ts from 'typescript';
+
+export class Rule extends Lint.Rules.AbstractRule {
+  public static metadata: Lint.IRuleMetadata = {
+    ruleName: 'no-conflicting-life-cycle-hooks',
+    type: 'maintainability',
+    description: 'Ensure that directives not implement conflicting lifecycle hooks.',
+    descriptionDetails: 'See more at https://angular.io/api/core/DoCheck#description.',
+    rationale: 'A directive typically should not use both DoCheck and OnChanges to respond ' +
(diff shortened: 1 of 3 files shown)


src/noConflictingLifeCycleHooksRule.ts
```diff
@@ -0,0 +1,72 @@
...
+    type: 'maintainability',
+    description: 'Ensure that directives not implement conflicting lifecycle hooks.',
+    descriptionDetails: 'See more at https://angular.io/api/core/DoCheck#description.',
+    rationale: 'A directive typically should not use both DoCheck and OnChanges to respond ' +
+      'to changes on the same input, as ngOnChanges will continue to be called when the ' +
+      'default change detector detects changes.',
+    options: null,
+    optionsDescription: 'Not configurable.',
+    typescriptOnly: true,
+  };
```
rafaelss95: Please, do not hesitate to suggest better descriptions :)

test/noConflictingLifeCycleHooksRule.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+      character: 8
+    }
+  }
+];
+
+describe(ruleName, () => {
+  describe('failure', () => {
+    it('should fail when implements both DoCheck and OnChanges hooks', () => {
+      const source = `
+        class Test implements DoCheck, OnChanges {
```
mgechev: TypeScript should fail with a compile-time error as well but I respect the completeness.

PR #566, opened 2018-04-21
chore: setup prettier

rafaelss95:
This:
- Removes conflicting tslint (stylistic) rules;
- Adds a few devDependencies:
  * "husky";
  * "lint-staged";
  * "prettier";
  * "rxjs-compat": Needed due to https://github.com/SamVerschueren/listr/issues/91 (`lint-staged` depends on this package);
- Formats all files using prettier;

See https://github.com/mgechev/codelyzer/pull/557#issuecomment-382574569.


rafaelss95:
Travis is failing atm and I'm not sure what would be the correct fix, but it seems like the `lint-staged` needs node >= 6 and travis is currently using node 5 and I also noticed that it's using `rxjs/zone.js` old versions.

mgechev:
Would you also switch to node stable from 5?

rafaelss95:
I've switched to the latest `LTS` and the error is different now:

> /Users/travis/.travis/job_stages: line 166: shell_session_update: command not found

wKoza:
@rafaelss95 , why did you add the devDependencies `husky` ? I don't see it used.

rafaelss95:
@wKoza [husky](https://github.com/typicode/husky) is used to manage git hooks. In this case, it'll run what I've defined in `lint-staged` (format and lint fixes) and also will prevent commits if the tests aren't passing.

wKoza:
Ok, thanks @rafaelss95. I'm in favor of adding a hook for the lint. BTW, I was afraid of having a hook that launches TUs. (sometimes, We want to push a TU in failure in order to fix it later).

PR #569, opened 2018-04-22
feat(rule): add option in max-inline-declarations to limit animations…

### LINKED ISSUE #568: [max-inline-declarations] - Add option for animations
Although not mentioned in the Angular style guide (maybe because it's a really new thing), I think it makes sense to warn in `max-inline-declarations` rule if the `animations` declaration is _too_ large as well as `template` and `styles`.

rafaelss95:
… lines

Closes #568.

### DIFF
--- a/src/maxInlineDeclarationsRule.ts
+++ b/src/maxInlineDeclarationsRule.ts
@@ -4,8 +4,10 @@ import { SourceFile } from 'typescript/lib/typescript';
 import { CodeWithSourceMap, ComponentMetadata } from './angular/metadata';
 import { NgWalker } from './angular/ngWalker';
 
+const DEFAULT_ANIMATIONS_LIMIT: number = 15;
 const DEFAULT_STYLES_LIMIT: number = 3;
 const DEFAULT_TEMPLATE_LIMIT: number = 3;
+const OPTION_ANIMATIONS = 'animations';
 const OPTION_STYLES = 'styles';
 const OPTION_TEMPLATE = 'template';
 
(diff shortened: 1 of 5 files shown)


src/maxInlineDeclarationsRule.ts
```diff
@@ -1,110 +1,130 @@
...
+
+export const stylesFailure = (limit: number): string => {
+  return generateFailure('styles', limit);
+};
+
+export const templateFailure = (limit: number): string => {
+  return generateFailure('template', limit);
+};
+
+export const animationsLinesDefaultLimit = 7;
```
rafaelss95: Not sure about this default limit...

wKoza:
This PR is confused because you are probably used `Prettier`. Two possibilities: 
- We can wait for your [`Prettier` PR](https://github.com/mgechev/codelyzer/pull/566) 
- You can revert changes in relation to 'Prettier'

rafaelss95:
Yes, since I've installed `Prettier` in my extensions, it always format when I save a file. Anyway, what is confusing for you?

mgechev:
@rafaelss95 I guess the big diff. Let's merge your PR which introduces prettier and after that merge this one.

mgechev:
@rafaelss95 would you give examples what the role of `transformAnimations` would be?

rafaelss95:
Sorry for the delayed response. To read animations parameter, I've followed what you did for `styles` and `template`. The intent of this PR is just to add another option for the existent rule (`max-inline-declarations`): limit animation lines. We could increase the default lines limit for animation.

src/maxInlineDeclarationsRule.ts
wKoza: It seems to me too little. Animation is quickly verbose. 15 lines, it is more common for a simple animation.

mgechev:
Let's see if we'll collect [enough votes](https://github.com/mgechev/codelyzer/issues?utf8=%E2%9C%93&q=label%3A%22votes+needed%22+sort%3Areactions-%2B1-desc+) for this feature before merging.

rafaelss95:
@mgechev So... let's close this?

rafaelss95:
bump @mgechev

Issue #573, opened 2018-04-24
Rule "no-life-cycle-call" does not allow for subclassing


jscharett:
If I extend a component and I override a lifecycle hook, this rule prevents me from calling the super method.  While I agree with the premise of the rule, it should, however, allow me to call the super method when overriding.

rafaelss95:
IMHO, this `rule` should prevent this as it does right now because any parent's lifecycle should be called by `Angular` itself. See https://github.com/angular/angular/issues/23091 and its related issues.

mgechev:
@jscharett makes sense to me. @rafaelss95 would you want to check if the receiver of the method call isn't `super`?

kevinphelps:
@rafaelss95: If I understand correctly, the issue you linked to is a bug where Angular does not call a lifecycle hook that is defined in a TypeScript mixin, not a base class.

Consider the following:

```
class BaseComponent {
  ngOnInit() {
    console.log('base component init');
  }
}

@Component({})
class MyComponent {
  ngOnInit() {
    super.ngOnInit();
    console.log('my component init');
  }
}
```

The `super.ngOnInit();` line should be allowed by the "no-lifecycle-call" rule because that is ONLY way `BaseComponent#ngOnInit` can be called in this case.

If you don't provide an implementation of `ngOnInit` in `MyComponent`, `BaseComponent#ngOnInit` would be invoked when Angular invokes `MyComponent#ngOnInit` simply because of how `extends` works.

But when you override a method, the outside world (i.e Angular) cannot cannot and should not call the super implementation.

rafaelss95:
You're totally right. I misread the issue. Good catch :)

kevinphelps:
If both classes are decorated, it works differently.

Issue #454, opened 2017-11-06
Build scripts are not cross-platform


reduckted:
The build scripts don't work on Windows. This makes it difficult for me (and others) to contribute to this project.

Some examples:

```
> yarn test
$ rimraf dist && tsc && cp -r test/fixtures dist/test && mocha dist/test --recursive
'cp' is not recognized as an internal or external command,
operable program or batch file.
error Command failed with exit code 1.
```

```
> yarn run prepare:package
$ cat package.json | ts-node build/package.ts > dist/package.json
'cat' is not recognized as an internal or external command,
operable program or batch file.
error Command failed with exit code 255.
```

It would be great if these scripts were cross-platform!

mgechev:
@reduckted I'll fix this as part of the next release. Will add it to my schedule for next week.

PR #575, opened 2018-04-24
fix(rule): fix bugs in no-life-cycle-call rule

kevinphelps:
- build: use 'copyfiles' instead 'cp' for building on Windows (cf49f4f7a7104de455bef344de28e01e02322b0a)

I can remove this commit if needed. The project would not build on my machine, so I replaced `cp` with `copyfiles` for cross-platform compatibility.

- fix(rule): allow super life cycle calls (0ad9abe418dd0a66dc4daf8946e323cfac17abea)

See #573. `super.ngOnInit()`, etc. should be allowed.

- fix(rule): prevent error when parent class node does not exist (dd63e7fa6a46b0f6be97e0c3e5b10e8f94394e02)

The `no-life-cycle-call` rule would error with a "cannot read name of undefined" error when the life cycle call was not in a class. In my case, there was test code calling `fixture.componentInstance.ngOnChanges()` and the rule would crash.


mgechev:
@kevinphelps thanks for the fix of https://github.com/mgechev/codelyzer/issues/454.

There seems to be a failure in Travis.

rafaelss95:
I think you should only fix the `no-life-cycle-call` on this PR and make a separate on for build changes.

kevinphelps:
I'm going to remove build change to see if it fixes the Travis build. It's in this PR simply because I needed for local development and I pushed it along with my fixes.

src/noLifeCycleCallRule.ts
```diff
@@ -14,7 +14,8 @@ export class Rule extends Lint.Rules.AbstractRule {
     typescriptOnly: true,
   };
 
-  static FAILURE_STRING: string = 'Avoid explicitly calls to lifecycle hooks in class "%s"';
+  static FAILURE_STRING_FOR_CLASS: string = 'Avoid explicit calls to lifecycle hooks in class "%s".';
+  static FAILURE_STRING: string = 'Avoid explicit calls to lifecycle hooks.';
```
rafaelss95: I think you can only have one `FAILURE_STRING` (and let's remove the unnecessary `:string`):

`static FAILURE_STRING = 'Avoid explicit calls to lifecycle hooks.';`

src/noLifeCycleCallRule.ts
```diff
@@ -49,19 +50,35 @@ export class ExpressionCallMetadataWalker extends NgWalker {
   }
 
   private validateCallExpression(node: ts.CallExpression): void {
-    const name = (node.expression as any).name;
+    const name = ts.isPropertyAccessExpression(node.expression) ? node.expression.name : undefined;
+    const expression = ts.isPropertyAccessExpression(node.expression) ? node.expression.expression : undefined;
```
rafaelss95: nit: extra new line.

src/noLifeCycleCallRule.ts
```diff
@@ -49,19 +50,35 @@ export class ExpressionCallMetadataWalker extends NgWalker {
...
+    const isSuperCall = expression && expression.kind === ts.SyntaxKind.SuperKeyword;
+    const isNotLifecycleCall = !name || !ts.isIdentifier(name) || !lifecycleHooksMethods.has(name.text as any);
+
+    if (isSuperCall || isNotLifecycleCall) {
       return;
     }
 
-    let currentNode = node as any;
+    let failureString = Rule.FAILURE_STRING;
+
```
rafaelss95: nit: extra new line.

src/noLifeCycleCallRule.ts
kevinphelps: You can have more than one.

rafaelss95: I didn't mention that it's impossible. I'm suggesting that you remove and keep only one failure. :)

kevinphelps: I can remove the type annotation. I kept it because it was already there.

kevinphelps: But I wanted to keep the class name when it's available.

src/noLifeCycleCallRule.ts
kevinphelps: I intentionally separated the if block with blank lines for readability.

rafaelss95: I don't mark the `if` block, but the block between `let` and `const` variables.

kevinphelps: I can remove if it violates the project's code conventions.

rafaelss95: Nah, it's just a nit. :)

src/noLifeCycleCallRule.ts
kevinphelps: I intentionally separated these variable declarations for readability. Source node info vs. rule logic checks.

src/noLifeCycleCallRule.ts
kevinphelps: I have updated this.

kevinphelps:
Thanks!

mgechev:
Would you look into the failing build? Looks prettier related.

kevinphelps:
Yeah, I can in a bit. Just need to reformat my changes since the prettier changes were merged first.

kevinphelps:
@mgechev: See #578.

PR #576, opened 2018-04-25
[WIP] feat(rule): add no-output-prefix rule

rafaelss95:
There's a rule to prevent prefixes on `@Output` (`no-output-on-prefix`) which is great, however it only prevents a single prefix: `on`, but people, depending on team could have another rules for prefixes and also what's considered a prefix (people may think that it shouldn't allow `onrequest` as it does for `onRequest`, for example, however it doesn't).

<hr>

This is similar to `no-input-prefix`, except that instead of passing an `array` of `strings`, you pass a string that'll be converted using `new RegExp` as [tslint#max-line-length#ignore-pattern](https://palantir.github.io/tslint/rules/max-line-length/) does to give more flexibility depending on what you consider a prefix or not.

For example, it will prevent this:

```
@Output() onPrefix = new EventEmitter<void>();
@Output() onprefix = new EventEmitter<void>();
@Output() yesPrefix = new EventEmitter<void>();
```

... if you define it like this:
`no-output-prefix: [true, '"^(on|yes)[a-zA-Z]+"']`


rafaelss95:
@mgechev I did mark this as WIP because I want to know what's your position about this.

mgechev:
@rafaelss95 although I like this PR, I haven't seen many feature requests for such feature.

If we decide to go with this rule, my proposal would be to drop `no-output-on-prefix` and generalize it with the current one, where the default regex is `/^on`.

rafaelss95:
Yes, the proposal was initially to replace the currently `no-output-on-prefix`. It surges on a team I'm working on as we needed to avoid other prefixes than just `on`.

Taking the fact that we're talking about `prefix`, what about accept regex in `no-input-prefix`?

mgechev:
Yes, makes sense for both rules.

PR #585, opened 2018-04-28
fix noInputRenameRule

### LINKED ISSUE #580: no-input-rename fails on input alias to directive selector
According to the changelog for version 4.3.0 the no-input-rename rule should now allow input aliasing for directives when the alias matches the selector name as described in the angular [attribute directive guide](https://angular.io/guide/attribute-directives#bind-to-an-input-alias). However after updating to 4.3.0 the following code is still reported as violating the no-input-rename rule:
```typescript
@Directive({
  selector: "[fssMsIconButtonLabel]"
})
export class IconButtonLabelDirective {
  @Input("fssMsIconButtonLabel") label: string;

  constructor(el: ElementRef) {}
}
```

wKoza:
In the hope that it clarifies things:

- A component shouldn't rename its @Input
- A directive shouldn't rename its @Input expect when the directive name is also an input property and the directive name doesn't describe the property.

Fixes #580

### DIFF
--- a/src/noInputRenameRule.ts
+++ b/src/noInputRenameRule.ts
@@ -16,7 +16,9 @@ export class Rule extends Lint.Rules.AbstractRule {
     typescriptOnly: true
   };
 
-  static FAILURE_STRING = 'In the class "%s", the directive input property "%s" should not be renamed.';
+  static FAILURE_STRING = 'In the class "%s", the directive input property "%s" should not be renamed. ' +
+  'However, you should use an alias when the directive name is also an input property, and the directive name' +
+  " doesn't describe the property. In this last case, you can disable this rule with `tslint:disable-next-line:no-input-rename`.";
 
   apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
     return this.applyWithWalker(new InputMetadataWalker(sourceFile, this.getOptions()));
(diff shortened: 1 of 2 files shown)


test/noInputRenameRule.spec.ts
```diff
@@ -56,19 +66,35 @@ describe(ruleName, () => {
...
           }
         `;
         assertAnnotated({
           ruleName,
-          message: getMessage('TestDirective', 'labelText'),
+          message: getMessage('TestDirective', 'label'),
+          source
+        });
+      });
+      it('should fail when a input property has the same name that the alias', () => {
```
rafaelss95: `an input`?

test/noInputRenameRule.spec.ts
```diff
@@ -56,19 +66,35 @@ describe(ruleName, () => {
         });
       });
 
-      it("should fail when input property is renamed and it's different from directive's selector", () => {
+      it('should fail when a input property has the same name that the alias', () => {
```
rafaelss95: `an input`?

src/noInputRenameRule.ts
```diff
@@ -34,7 +36,11 @@ export class InputMetadataWalker extends NgWalker {
     const className = (property.parent as ts.PropertyAccessExpression).name.getText();
     const memberName = property.name.getText();
 
-    if (args.length === 0 || (this.directiveSelector && this.directiveSelector.indexOf(memberName) !== -1)) {
+    if (
+      args.length === 0 ||
+      (this.directiveSelector &&
+        (input.expression as any).arguments.some(arg => this.directiveSelector.indexOf(arg.text) !== -1 && memberName !== arg.text))
```
rafaelss95: Instead of casting to `any`, what about `(input.expression as ts.CallExpression)`?

test/noInputRenameRule.spec.ts
```diff
@@ -3,18 +3,24 @@ import { assertSuccess, assertAnnotated } from './testHelper';
 const ruleName = 'no-input-rename';
 
 const getMessage = (className: string, propertyName: string): string => {
-  return `In the class "${className}", the directive input property "${propertyName}" should not be renamed.`;
+  return (
```
rafaelss95: I'd refactor it to:

```
const { FAILURE_STRING, metadata: { ruleName } } = Rule;

const getFailureMessage = (className: string, propertyName: string): string => {
  sprintf(FAILURE_STRING, className, propertyName);
};
```

PR #586, opened 2018-04-28
feat(rule): add prefer-inline-decorator rule

### LINKED ISSUE #549: [Feature Request] Enforce single line @Input decorator
I'd like to have a rule to enforce the recommendation of 'Consider placing @Input() or @Output() on the same line as the property it decorates.' described in Style 05-12.  As far as I can tell use-input-property-decorator only enforces the usage of @Input not the styling

rafaelss95:
This ensures that decorators are on the same line as the property it decorates.

Failure example:
```
@Input()
x: string;
```

Success example:
```
@Input() x: string;
```

Closes #549.

**Note that I've tried to implement a fix for this.** Let me know if I did something wrong.

### DIFF
--- a/src/preferInlineDecoratorRule.ts
+++ b/src/preferInlineDecoratorRule.ts
@@ -1,0 +1,91 @@
+import { IOptions, IRuleMetadata, Replacement, RuleFailure, Rules } from 'tslint/lib';
+import { isSameLine } from 'tsutils';
+import { Decorator, Node, PropertyAccessExpression, SourceFile } from 'typescript';
+
+import { NgWalker } from './angular/ngWalker';
+import { getDecoratorName } from './util/utils';
+
+export class Rule extends Rules.AbstractRule {
+  static readonly metadata: IRuleMetadata = {
+    description: 'Ensures that decorators are on the same line as the property/method it decorates.',
+    descriptionDetails: 'See more at https://angular.io/guide/styleguide#style-05-12.',
+    hasFix: true,
+    optionExamples: ['true', '[true, "HostListener"]'],
+    options: {
+      items: {
+        type: 'string'
+      },
+      type: 'array'
+    },
+    optionsDescription: 'A list of blacklisted decorators.',
+    rationale: 'Placing the decorator on the same line usually makes for shorter code and still easily identifies the property/method.',
+    ruleName: 'prefer-inline-decorator',
+    type: 'style',
+    typescriptOnly: true
+  };
+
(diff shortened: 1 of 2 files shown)


rafaelss95:
@mgechev:

I've implemented it for all "property" decorators, does it makes sense to add it to `@HostListener`, for example?

It would fail in this case:
```
@HostListener('keydown')
newColor() {
  // something nice here
}
```

... but not for this:
```
@HostListener('keydown') newColor() {
  // something nice here
}
```

<hr>

I've tried to verify the case I described in https://github.com/mgechev/codelyzer/issues/549#issuecomment-384494828, however it seems that this case does not trigger either the `visitPropertyDecorator` method or the` visitNgInput` method, so I'm stucked.

wKoza:
sure, implement this for method make sense. @HostListener is good candidate.
Concerning the configuration of the rule, the global approach seems the best way since we want to identify easily the role of an input / method.

src/preferDecoratorInlineRule.ts
```diff
@@ -0,0 +1,44 @@
...
+export class Rule extends Rules.AbstractRule {
+  static readonly metadata: IRuleMetadata = {
+    description: 'Ensures that decorators are on the same line as the property it decorates.',
+    descriptionDetails: 'See more at https://angular.io/guide/styleguide#style-05-12.',
+    hasFix: true,
+    options: null,
+    optionsDescription: 'Not configurable.',
+    rationale:
+      'Placing the decorator on the same line usually makes for shorter code and still easily identifies the property as an input or output.',
+    ruleName: 'prefer-decorator-inline',
```
mgechev: I'd call it `prefer-inline-decorator`.

src/preferDecoratorInlineRule.ts
```diff
@@ -0,0 +1,44 @@
+import { IRuleMetadata, Replacement, RuleFailure, Rules } from 'tslint';
+import { isSameLine } from 'tsutils';
+import { Decorator, Node, PropertyAccessExpression, SourceFile } from 'typescript';
+import { NgWalker } from './angular/ngWalker';
+
+export class Rule extends Rules.AbstractRule {
```
mgechev: It makes sense to allow rule configuration (i.e.blacklist decorators for example).

rafaelss95: Done.

src/preferDecoratorInlineRule.ts
rafaelss95: Done.

PR #587, opened 2018-04-29
refactor: replace/remove deprecated method calls

rafaelss95:
This:

- Adds `deprecation` rule to the `tslint.json`;
- Replaces all usages of deprecated methods (in majority `tslint` methods);

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "format:base": "prettier --config ./.prettierrc \"*.{json,md}\" \"src/**/*.{css,scss,ts}\" \"test/**/*.{css,scss,ts}\"",
     "format:check": "npm run format:base -- --list-different",
     "format:fix": "npm run format:base -- --write",
-    "lint": "tslint -c tslint.json \"src/**/*.ts\" \"test/**/*.ts\"",
+    "lint": "tslint -p . -c tslint.json \"src/**/*.ts\" \"test/**/*.ts\"",
     "lint:fix": "npm run lint -- --fix",
     "release": "npm run build && rimraf dist && tsc -p tsconfig-release.json && npm run copy:common && npm run prepare:package && cross-env BUILD_TYPE=prod npm run set:vars",
     "build": "rimraf dist && tsc && npm run format:check && npm run lint && npm t",
--- a/src/angular/sourceMappingVisitor.ts
+++ b/src/angular/sourceMappingVisitor.ts
@@ -104,6 +104,7 @@ export class SourceMappingVisitor extends RuleWalker {
 
   createFailure(s: number, l: number, message: string, fix?: Fix): RuleFailure {
     const { start, length } = this.getMappedInterval(s, l);
+    // tslint:disable-next-line:deprecation
     return super.createFailure(start, length, message, fix);
   }
 
--- a/src/templatesNoNegatedAsyncRule.ts
+++ b/src/templatesNoNegatedAsyncRule.ts
@@ -16,18 +16,21 @@ class TemplateToNgTemplateVisitor extends RecursiveAngularExpressionVisitor {
       return super.visitBinary(expr, context);
     }
 
-    const operator = this.codeWithMap.code.slice(expr.left.span.end, expr.right.span.start);
+    const {
+      left: {
+        span: { end: endLeftSpan }
+      },
+      right: {
+        span: { start: startRightSpan }
+      },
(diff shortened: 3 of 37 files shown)


src/useLifeCycleInterfaceRule.ts
```diff
@@ -61,28 +61,21 @@ export class ClassMetadataWalker extends Lint.RuleWalker {
     return interfaces;
   }
 
-  private validateMethods(methods: any[], interfaces: string[], className: string) {
+  private validateMethods(methods: ts.ClassElement[], interfaces: string[], className: string) {
     methods.forEach(m => {
-      let n = (<any>m.name).text;
-      if (n && this.isMethodValidHook(m, interfaces)) {
-        let hookName = n.substr(2, n.lenght);
```
rafaelss95: Fun fact: `lengHt` was mispelled (2 times in this class) :)

mgechev: Oh!

mgechev:
@rafaelss95 thanks for the cleanup! Would you check why the build is failing?

rafaelss95:
@mgechev so... that was what I asked above, if I change in these 2 places I mentioned, the tests break, I really don't know why.

mgechev:
```
ERROR: /Users/travis/build/mgechev/codelyzer/src/angular/sourceMappingVisitor.ts[107, 18]: createFailure is deprecated: Prefer `addFailureAt` and its variants.
ERROR: /Users/travis/build/mgechev/codelyzer/src/angular/templates/basicTemplateAstVisitor.ts[186, 53]: addFailure is deprecated: Prefer `addFailureAt` and its variants.
```

I guess we should not use deprecated methods.

rafaelss95:
I think you don't understand what I mean. I'm trying to say that if I change these deprecated calls it break some tests. See https://github.com/mgechev/codelyzer/pull/587#issuecomment-385220278

rafaelss95:
@mgechev I've pushed the changes for those leftovers. Now you can see the full log error in travis.

rafaelss95:
bump @mgechev.

mgechev:
@rafaelss95 will take a look in a few minutes.

package.json
```diff
@@ -8,7 +8,7 @@
     "format:base": "prettier --config ./.prettierrc \"*.{json,md}\" \"src/**/*.{css,scss,ts}\" \"test/**/*.{css,scss,ts}\"",
     "format:check": "npm run format:base -- --list-different",
     "format:fix": "npm run format:base -- --write",
-    "lint": "tslint -c tslint.json \"src/**/*.ts\" \"test/**/*.ts\"",
+    "lint": "tslint -p . -c tslint.json \"src/**/*.ts\" \"test/**/*.ts\"",
```
mgechev: Do we need `-p`? Are we using any rules with type checking?

src/angular/sourceMappingVisitor.ts
```diff
@@ -104,6 +104,7 @@ export class SourceMappingVisitor extends RuleWalker {
 
   createFailure(s: number, l: number, message: string, fix?: Fix): RuleFailure {
     const { start, length } = this.getMappedInterval(s, l);
+    // tslint:disable-next-line:deprecation
     return super.createFailure(start, length, message, fix);
```
mgechev: `createFailure` is deprecated?

package.json
mgechev: I see that's required because of the `deprecation` rule?

src/angular/sourceMappingVisitor.ts
rafaelss95: Yep, it has been deprecated for over a year. PR https://github.com/palantir/tslint/pull/1952

package.json
rafaelss95: Yes, `deprecation` rule requires type information.

src/templatesNoNegatedAsyncRule.ts
```diff
@@ -16,39 +16,41 @@ class TemplateToNgTemplateVisitor extends RecursiveAngularExpressionVisitor {
       return super.visitBinary(expr, context);
     }
 
-    const operator = this.codeWithMap.code.slice(expr.left.span.end, expr.right.span.start);
+    const {
```
mgechev: Let's use less of these assignments. A bit hard to follow :-)

Issue #519, opened 2018-02-19
Why is the pipe-naming rule deprecated?


yannickglt:
Hey!
Sorry to create an issue to ask such a question but I could not find any answer on the repo or in the doc.
I saw that the `pipe-naming` rule was deprecated in the Readme. Could you tell me what's the reason and if it was replaced with another rule?
Thanks

mgechev:
You can name your pipes only camel case if you try to use snake-case then your application will not compile. This is a compile-time error, not a linting one. The alternative is to use AoT.

yannickglt:
Thanks for your quick answer!

Good to learn that there's a compile error now. 

I mainly wanted to know if the practices about prefixing pipes have changed. If it is not a good practice anymore then I guess the rule is not useful anymore.

mgechev:
It's still considered a good practice. Prefixing will not be caught by the compiler thought.

yannickglt:
Great. Thank you!

emilio-martinez:
@mgechev is there a replacement for checking that the Pipe names are prefixed accordingly?

```
"pipe-naming": [true, "camelCase", "app"]
                                     ^
```

georgms:
Sorry for resurrecting this but I can still do:

```typescript
@Pipe({ name: 'underscore_name' })
export class UnderscoreNamePipe implements PipeTransform {}
```

But according to the Angular style guide [pipe name strings should use *lowerCamelCase*](https://angular.io/guide/styleguide#pipe-names), ie. `@Pipe({ name: 'underscoreName' })`.

How can I enforce *lowerCamelCase* for pipe names?

mgechev:
I'd recommend opening an issue in [angular-eslint](https://github.com/angular-eslint/angular-eslint). Codelyzer is now deprecated.

PR #593, opened 2018-05-01
chore: add deprecation message to 'pipe-naming' rule

rafaelss95:
Once someone try to use this _deprecated_ rule, it'll show a warning like:

> pipe-naming is deprecated. Message here

<hr>

Not sure if it's the best message, but I've picked it in https://github.com/mgechev/codelyzer/issues/519#issuecomment-366785640.

### DIFF
--- a/src/pipeNamingRule.ts
+++ b/src/pipeNamingRule.ts
@@ -8,6 +8,7 @@ import { IOptions } from 'tslint';
 
 export class Rule extends Lint.Rules.AbstractRule {
   public static metadata: Lint.IRuleMetadata = {
+    deprecationMessage: 'You can name your pipes only camelCase. If you try to use snake-case then your application will not compile.',
     ruleName: 'pipe-naming',
     type: 'style',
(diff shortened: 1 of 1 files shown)


rafaelss95:
@mgechev @wKoza while I do understand that it's deprecated because of the `snake-case` part, I can still see useful things in it such as enforcing prefixes like this:

```
{
  "rules": {
    "pipe-naming": [true, "^ngx?[A-Z]+"]
    // or less generic, as an array
    "pipe-naming": [true, "ng", "ngx"]
  }
}
```

Can't we just deprecate the actual configuration?

mgechev:
@rafaelss95 I like the regex config:

```
"pipe-naming": [true, "^ngx?[A-Z]+"]
```

It'll make more sense to rename the rule. If we have a rule with the same name but different semantics we'll create a lot of confusion.

rafaelss95:
Does it makes sense to create another rule now? Or better wait for complete removal of this?

Issue #594, opened 2018-05-02
About prefixes rules


rafaelss95:
In theory, `prefix` is a word, letter, or number placed before another, however I noticed that there's a test case in `noOutputOnPrefix`:

https://github.com/mgechev/codelyzer/blob/651be15e7a5e0ce047b1ab0be5a90b51a3adb4aa/test/noOutputOnPrefixRule.spec.ts#L35-L47

... where it implies that we can't name a variable `on`. Is it really the expected behavior? For me, it's a little strange because there's nothing after `on` word.

Anyway, I'm asking because actually `noInputPrefix` works a little bit different. In which rule should go a fix?

mgechev:
To me, it sounds fine to disallow `on` as the name of an output. Let's reopen this if anyone disagrees.

PR #597, opened 2018-05-03
fix(no-input-prefix): exact strings not being reported

rafaelss95:
Fixed according to discussion in #594.

### DIFF
--- a/test/noInputPrefixRule.spec.ts
+++ b/test/noInputPrefixRule.spec.ts
@@ -5,89 +5,93 @@ const {
   FAILURE_STRING,
   metadata: { ruleName }
 } = Rule;
-const className = 'Test';
 
-const getFailureAnnotations = (num: number): string => {
-  return '~'.repeat(num);
-};
-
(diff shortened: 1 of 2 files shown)


test/noInputPrefixRule.spec.ts
```diff
@@ -36,6 +36,25 @@ describe(ruleName, () => {
...
     });
 
+    it('should fail when an input property is strictly equal to a blacklisted prefix', () => {
+      const prefixes = ['should'];
+      const propertyName = prefixes[0];
+      const inputExpression = `@Input() ${propertyName}: boolean;`;
+      const source = `
+        @Directive()
+        class ${className} {
+          ${inputExpression}
```
mgechev: Let's not use interpolation that aggressively. If we duplicate `should` twice in a single test case, this is fine. Currently, the code is a bit hard to follow.

rafaelss95: Right. Changing...

rafaelss95: Done. I hope it's clear now :)

PR #603, opened 2018-05-05
docs: fix inconsistent descriptions

rafaelss95:
This normalize (and fix, in some cases) all rule's description, failure and rationale.


mgechev:
That's a great PR @rafaelss95! Would you see why's Travis complaining?

PR #604, opened 2018-05-05
fix: add missing super calls

rafaelss95:


rafaelss95:
I'm not sure why a test case in `angularWhitespaceRule` is failing on `appveyor`. Works both on my machine and on travis.

PR #626, opened 2018-05-12
docs: remove option examples deprecated usage

rafaelss95:
PR https://github.com/palantir/tslint/pull/2527 deprecated passing an `array` of strings for `optionExamples`.

In addition to this, I've refactored the metadata in general to be more generic as `tslint` does, using `const` variables for `options` and so on.

### DIFF
--- a/src/i18nRule.ts
+++ b/src/i18nRule.ts
@@ -1,13 +1,16 @@
+import * as ast from '@angular/compiler';
 import * as Lint from 'tslint';
 import * as ts from 'typescript';
 import { NgWalker } from './angular/ngWalker';
-import * as ast from '@angular/compiler';
 import { BasicTemplateAstVisitor } from './angular/templates/basicTemplateAstVisitor';
 
-type Option = 'check-id' | 'check-text';
+const OPTION_CHECK_ID = 'check-id';
+const OPTION_CHECK_TEXT = 'check-text';
+
+type CheckOption = typeof OPTION_CHECK_ID | typeof OPTION_CHECK_TEXT;
 
 interface ConfigurableVisitor {
-  getOption(): Option;
+  getCheckOption(): CheckOption;
 }
 
 class I18NAttrVisitor extends BasicTemplateAstVisitor implements ConfigurableVisitor {
@@ -32,7 +35,7 @@ class I18NAttrVisitor extends BasicTemplateAstVisitor implements ConfigurableVis
     super.visitAttr(attr, context);
   }
 
-  getOption(): Option {
+  getCheckOption(): CheckOption {
     return 'check-id';
   }
 }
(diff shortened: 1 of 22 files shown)


rafaelss95:
@wKoza could you review?

src/i18nRule.ts
```diff
@@ -135,31 +138,30 @@ class I18NTemplateVisitor extends BasicTemplateAstVisitor {
...
+  static readonly metadata: Lint.IRuleMetadata = {
     description: 'Ensures following best practices for i18n.',
-    rationale: 'Makes the code more maintainable in i18n sense.',
-    optionsDescription: Lint.Utils.dedent`
-      Arguments may be optionally provided:
-      * \`"check-id"\` Makes sure i18n attributes have ID specified
-      * \`"check-text"\` Makes sure there are no elements with text content but no i18n attribute
-    `,
-
+    optionExamples: [true, [true, OPTION_CHECK_ID]],
```
wKoza: What's happen when we use only `true` ?

src/i18nRule.ts
```diff
@@ -135,31 +138,30 @@ class I18NTemplateVisitor extends BasicTemplateAstVisitor {
...
+        type: 'string'
       },
+      maxLength: 2,
       minLength: 0,
-      maxLength: 3
+      type: 'array'
     },
-    optionExamples: ['[true, "check-id"]'],
+    optionsDescription: Lint.Utils.dedent`
+      Arguments may be optionally provided:
```
wKoza: is it really Optional ?

src/preferInlineDecoratorRule.ts
```diff
@@ -1,20 +1,39 @@
...
+
+export const decoratorKeys: ReadonlySet<string> = new Set(enumKeys);
+
 export class Rule extends Rules.AbstractRule {
   static readonly metadata: IRuleMetadata = {
     description: 'Ensures that decorators are on the same line as the property/method it decorates.',
     descriptionDetails: 'See more at https://angular.io/guide/styleguide#style-05-12.',
     hasFix: true,
-    optionExamples: ['true', '[true, "HostListener"]'],
+    optionExamples: [true, [true, Decorators.HostListener]],
```
wKoza: What's happen when we use only `true` ? Does that make sense ?

rafaelss95: It'll report a warn for every decorator. The list is a blacklist. In this example, you'll not receive any failure about `@HostListener`.

src/i18nRule.ts
rafaelss95: Fixed.

rafaelss95:
Ready for another round :rocket:

src/maxInlineDeclarationsRule.ts
```diff
@@ -47,6 +47,17 @@ export class Rule extends Rules.AbstractRule {
...
+
+  isEnabled(): boolean {
+    const {
+      metadata: {
+        options: { maxLength, minLength }
+      }
+    } = Rule;
+    const { length } = this.ruleArguments;
+
+    return super.isEnabled() && length >= minLength && length <= maxLength;
```
rafaelss95: I think we should disable this rule also, if either the `styles` or `template` are negative values.

PR #631, opened 2018-05-26
chore: enable strict

### LINKED ISSUE #629: Turn on strict TS compilation mode
We started syncing this repo for google internal usage. Technically we have exceptions for code that doesn't have `'strict': true`, but it is generally best to use that flag to be consistent with the rest of google TS users.

rafaelss95:
Closes #629.


mgechev:
@rafaelss95 🤘

tsconfig.json
```diff
@@ -8,6 +8,8 @@
     "noLib": false,
     "outDir": "./dist",
     "removeComments": true,
+    "strict": true,
+    "strictPropertyInitialization": false,
```
mgechev: @rkirov does it work for you if we disable `strictPropertyInitialization`?

mgechev:
@rafaelss95 great job! LGTM, I'll let @wKoza to have a look as well.

tsconfig.json
rafaelss95: @mgechev There are only few some places that forced me to disable this. I can enable this and initialize properties properly or even use a `!` to indicate that I want _hack_ that property. Let me know what should I do.

rafaelss95:
Thanks, @mgechev :)

rafaelss95:
Ping @wKoza.

tsconfig.json
wKoza: If the job isn't too hard, I would prefer use the default value of `strictPropertyInitialization`. Could you update this PR in this direction ?

rafaelss95: @wKoza Updated.

Issue #657, opened 2018-06-02
export NgWalker for library users


piyukore06:
Hello Maintainers,

Thank you for building Codelyzer. It is really helpful not just for catching lint errors but also knowing what is a good way of doing a certain thing.
 
TsLint has exported it's Walkers. On top of which devs can build custom TsLint rules. I wanted to build a custom rule using angular metadata but I found out that NgWalker is not exported.

Is there any particular reason it was not exported?
I am really new at this so if there is any other way of doing the same, pls let me know.

mgechev:
Hey @piyukore06, I've thought of externalizing the `NgWalker` several times. Your proposal makes a lot of sense. To prioritize this feature request, I'll add `votes needed` label and close it, so that people can vote and we can introduce it as part of the roadmap if the demand is high. Alternatively, if you have some time, we'd love to get a PR from you!

rafaelss95:
@mgechev Let's remove the 'votes needed` label from issues that have already been finalized? It's a little confusing...

PR #658, opened 2018-06-02
feat: externalize ngWalker to support custom lint rules

piyukore06:
relates to #657

### DIFF
--- a/src/angular/index.ts
+++ b/src/angular/index.ts
@@ -1,0 +1,7 @@
+export * from './templates/basicTemplateAstVisitor';
+
+export * from './styles/basicCssAstVisitor';
+
+export * from './config';
+export * from './metadata';
+export * from './ngWalker';
--- a/src/index.ts
+++ b/src/index.ts
@@ -37,7 +37,7 @@ export { Rule as UsePipeDecoratorRule } from './usePipeDecoratorRule';
 export { Rule as UsePipeTransformInterfaceRule } from './usePipeTransformInterfaceRule';
 export { Rule as UseViewEncapsulationRule } from './useViewEncapsulationRule';
 
-export * from './angular/config';
+export * from './angular';
 
 // this file exists for tslint to resolve the rules directory
 export const rulesDirectory = '.';


mgechev:
Thanks for the quick PR! My idea was to externalize the content of the `angular` directory into a separate package. Does this cover your use case completely? Keep in mind that there are more visitor abstractions for traversal of template, style, and expression ASTs.

piyukore06:
Hi @mgechev, Thank you for your input. I only needed NgWalker for my use case, but I think it makes sense to export the angular directory contents, there's so much goodness, everyone could use to build rules.
Just one question, When you are saying a separate package, do you mean npm module like @codelyzer/angular and @codelyzer/rules something like that?

piyukore06:
Yea I am trying but couldn't find anything.. locally the command runs without any problems.

src/angular/index.ts
```diff
@@ -0,0 +1,21 @@
+export * from './templates/basicTemplateAstVisitor';
+export * from './templates/jitReflector';
```
mgechev: I don't think we should export the `jitReflector` as part of the public API.

src/angular/index.ts
```diff
@@ -0,0 +1,21 @@
...
+export * from './templates/recursiveAngularExpressionVisitor';
+export * from './templates/referenceCollectorVisitor';
```
mgechev: Better not export `referenceCollectorVisitor`.

src/angular/index.ts
```diff
@@ -0,0 +1,21 @@
...
+export * from './templates/templateParser';
+
+export * from './styles/basicCssAstVisitor';
```
mgechev: Can we keep only the `basicCssAstVisitor` exported?

src/angular/index.ts
```diff
@@ -0,0 +1,21 @@
...
+export * from './styles/cssAst';
+export * from './styles/cssLexer';
+export * from './styles/cssParser';
+export * from './styles/parseCss';
+export * from './styles/parseUtil';
+
+export * from './config';
+export * from './expressionTypes';
+export * from './metadata';
+export * from './metadataReader';
```
mgechev: Can we skip the `metadataReader`?

src/angular/index.ts
```diff
@@ -0,0 +1,21 @@
...
+export * from './ngWalker';
+export * from './ngWalkerFactoryUtils';
+export * from './sourceMappingVisitor';
```
mgechev: My guess is that the error comes from the missing newline.

mgechev:
@piyukore06 if TypeScript won't complain, I'd suggest exporting only the base visitors without any extra internal APIs. This way, we'd be able to make changes in them without introducing issues for consumers of codelyzer.

piyukore06:
@mgechev thank you for your comments! They lead me to a better understanding of, where to look and what should be exported.

PR #659, opened 2018-06-02
refactor: use type guard instead of unnecessary or incorrect type assertions

rafaelss95:
This:

- Replaces a lot of **any** usages with more expressive types;
- Uses type guard instead of (unnecessary) type casts;

PS: You can see that there are a lot more places that could be changed, but these or I can't replace, because changing them, breaks something or unfortunately it has to be **any**.

### DIFF
--- a/src/angular/ngWalker.ts
+++ b/src/angular/ngWalker.ts
@@ -15,11 +15,22 @@ import { RecursiveAngularExpressionVisitor } from './templates/recursiveAngularE
 import { ReferenceCollectorVisitor } from './templates/referenceCollectorVisitor';
 import { parseTemplate } from './templates/templateParser';
 
-const getDecoratorStringArgs = (decorator: ts.Decorator): (string | null)[] => {
+const getDecoratorStringArgs = (decorator: ts.Decorator): string[] => {
   const { expression } = decorator;
   const args = ts.isCallExpression(expression) ? expression.arguments : ts.createNodeArray();
 
-  return args.map(a => (ts.isStringLiteral(a) ? a.text : null));
+  return args.filter(ts.isStringLiteral).map(x => x.text);
+};
+
+const getPosition = (node: ts.Node) => {
+  let pos = 0;
+  if (node) {
+    pos = node.pos + 1;
+    try {
+      pos = node.getStart() + 1;
+    } catch {}
+  }
+  return pos;
 };
 
 export interface NgWalkerConfig {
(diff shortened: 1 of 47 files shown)


src/angular/ngWalker.ts
```diff
@@ -107,56 +118,48 @@ export class NgWalker extends Lint.RuleWalker {
...
     for (const style of styles) {
       try {
-        this.visitNgStyleHelper(parseCss(style.style.code), metadata, style, getPosition(style.node!));
+        const cssAst = parseCss(style.style.code);
+        this.visitNgStyleHelper(cssAst, metadata, style, getPosition(style.node!));
       } catch (e) {
-        logger.error('Cannot parse the styles of', ((<any>metadata.controller || {}).name || {}).text, e);
+        const {
+          controller: { name }
+        } = metadata;
```
lazarljubenovic: This needs defaults in order to guard the same way the original did. Not sure if intentional.

rafaelss95: The _property_ `controller` always exists according to:

https://github.com/mgechev/codelyzer/blob/da0f553fe7d580f38907f9baa9124cbc55e37c96/src/angular/metadata.ts#L23-L27

... and it makes sense, because once we instantiate a `DirectiveMetadata`, we pass all properties:

https://github.com/mgechev/codelyzer/blob/da0f553fe7d580f38907f9baa9124cbc55e37c96/src/angular/metadataReader.ts#L63-L67

... now I'm thinking: why do we need to use `Object.assign` instead of a trivial `constructor`?

<hr>

Anyway, these "defaults" should have been changed in #631, but I really forgot about it. :disappointed:

rafaelss95:
bump :alien:

src/angular/sourceMappingVisitor.ts
```diff
@@ -95,7 +95,7 @@ export class SourceMappingVisitor extends RuleWalker {
   parentAST: any;
   private consumer: SourceMapConsumer;
 
-  constructor(sourceFile: ts.SourceFile, options: IOptions, protected codeWithMap: CodeWithSourceMap, protected basePosition: number) {
+  constructor(sourceFile: ts.SourceFile, options: IOptions, public codeWithMap: CodeWithSourceMap, protected basePosition: number) {
```
mgechev: Why not `protected`?

src/angular/templates/basicTemplateAstVisitor.ts
```diff
@@ -8,7 +8,7 @@ import { ComponentMetadata } from '../metadata';
 import { RecursiveAngularExpressionVisitor } from './recursiveAngularExpressionVisitor';
 import { SourceMappingVisitor } from '../sourceMappingVisitor';
 
-const getExpressionDisplacement = (binding: any) => {
+const getExpressionDisplacement = (binding: ast.TemplateAst) => {
```
mgechev: ❤️

src/angular/sourceMappingVisitor.ts
rafaelss95: It was necessary because I've changed the type of `context` in `trackByRule#35` from `any` to `BasicTemplateAstVisitor` (https://github.com/mgechev/codelyzer/pull/659/files#diff-26d9296d5e0a25d3fe6a9b174e67da35R35)... it gives an error with `protected`:

> src/trackByFunctionRule.ts(44,30): error TS2446: Property 'codeWithMap' is protected and only accessible through an instance of class 'TrackByFunctionTemplateVisitor'.

PR #674, opened 2018-06-24
fix(no-output-on-prefix): fix regular expression

loktionov129:
this commit fixes incorrect behavior when memberName contains 'on' suffix in the word, for example selectionChanged.

```
/on((?![a-z])|(?=$))/.test('selectionChanged');
// true

/^on((?![a-z])|(?=$))/.test('selectionChanged');
// false
```

### DIFF
--- a/test/noOutputOnPrefixRule.spec.ts
+++ b/test/noOutputOnPrefixRule.spec.ts
@@ -69,4 +69,16 @@ describe('no-output-on-prefix', () => {
       assertSuccess('no-output-on-prefix', source);
     });
   });
+
+  describe('valid output property name', () => {
+    it("should succeed, when a output property containing 'on' suffix", () => {
+      const source = `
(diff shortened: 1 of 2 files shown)


rafaelss95:
Good catch! Could you add a test containing a case like this you mentioned?

test/noOutputOnPrefixRule.spec.ts
```diff
@@ -69,4 +69,16 @@ describe('no-output-on-prefix', () => {
       assertSuccess('no-output-on-prefix', source);
     });
   });
+
+  describe('valid output property name', () => {
+    it("should succeed, when a output property containing 'on' suffix", () => {
```
rafaelss95: `a` -> `an`. Also, you can put this test in the above `describe`.

rafaelss95:
LGTM!

PR #675, opened 2018-06-24
build: add standard-version and commitlint packages

rafaelss95:
Adds:

- [standard-version](https://github.com/conventional-changelog/standard-version) - a package to generate changelog following specific (in our case `Angular`) conventions and tags;
- [commitlint](https://github.com/marionebl/commitlint) - a package to validate commit messages (it will really improve the changelog generator). It was configured according to [CONTRIBUTING.md](https://github.com/mgechev/codelyzer/blob/master/CONTRIBUTING.md);


loktionov129:
@rafaelss95,
I think this package https://github.com/okonet/lint-staged would be great too for running linters on git staged files.

And it seems to me https://github.com/semantic-release/semantic-release are better than https://github.com/conventional-changelog/standard-version

rafaelss95:
@loktionov129 Thanks for your suggestions.

We already use `lint-staged` for running tslint and prettier, however `commitlint` has another purpose: validate commit messages to generate a good/useful changelog.

While I do like the idea of `semantic-release`, I don't think it's a good idea to add it here, because not every time we accept a PR, we want to make a release.

PR #682, opened 2018-06-27
Fix typo of length closes #681

### LINKED ISSUE #681: Property length misspelled twice in useLifeCycleInterfaceRule.js
Hi,
I found that in [useLifeCycleInterfaceRule.js](https://github.com/mgechev/codelyzer/blob/master/docs/dist/src/useLifeCycleInterfaceRule.js)  [L#83](https://github.com/mgechev/codelyzer/blob/master/docs/dist/src/useLifeCycleInterfaceRule.js#L83) and [L#91](https://github.com/mgechev/codelyzer/blob/master/docs/dist/src/useLifeCycleInterfaceRule.js#L91) contain the property `.lengHT` misspelled two times. This was already mentioned [here](https://github.com/mgechev/codelyzer/pull/587#discussion_r184898534).
Just to keep track!
Cheers!

robzenn92:
This is related to #681

### DIFF
--- a/docs/dist/src/useLifeCycleInterfaceRule.js
+++ b/docs/dist/src/useLifeCycleInterfaceRule.js
@@ -80,7 +80,7 @@ var ClassMetadataWalker = (function (_super) {
         methods.forEach(function (m) {
             var n = m.name.text;
             if (n && _this.isMethodValidHook(m, interfaces)) {
-                var hookName = n.substr(2, n.lenght);
+                var hookName = n.substr(2, n.length);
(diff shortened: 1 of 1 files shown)


rafaelss95:
I guess the right fix would be generating docs up to date with latest version..

mgechev:
Yes, @rafaelss95 I agree. I still believe we can merge this now and regenerate the docs as well.

rafaelss95:
@mgechev can you tell me the steps to do this? I noticed that there's a script called `buildDocs` (which isn't working at the moment) that generates HTML docs, however what about all other files in docs folder?

PR #686, opened 2018-07-03
fix: url not passed to transformStyle when using styleUrls

alexkpek:
Fixed bug when url is not passed to transformStyle function when using styleUrls.


mgechev:
Thanks for the PR!

Issue #307, opened 2017-05-11
Allow components/directives to use both element and attribute as selectors


diagramatics:
Right now `directive-selector` and `component-selector` chooses between using `attribute` or `element` as the type, but you can't actually use both. I'd like to make a case for supporting having the two if it's not possible yet.

After listening to [Angular Material team's talk in ng-conf about component API designs](https://www.youtube.com/watch?v=0BikjL858OQ) there is a case for using attribute selectors as components to allow components to be able to use the DOM API without needing to reflect 20+ bindings from the custom element style components to the inner DOM API it wraps. angular/material2 itself uses directives — which I'm not sure why directives instead of components — my case for using components would be for the scoped styling it provides compared to directives, which angular/material2 doesn't need to yet with their approach to having a global stylesheet with all the styles of the components.

I don't have a use case example for directives yet though, but considering the configuration of both rules are the same, maybe it's worth giving people the option to anyway.

mgechev:
It's already possible to pass an array with elements `["element", "attribute"]` for both. Another alternative is to disable the rules.

diagramatics:
Ah my bad, thanks for the clarification.

PR #698, opened 2018-08-08
docs: component-selector both element and attribute rule

PapsOu:
Related to bug #307


src/componentSelectorRule.ts
```diff
@@ -46,7 +47,7 @@ export class Rule extends SelectorRule {
     },
     optionsDescription: Utils.dedent`
       Options accept three obligatory items as an array:
-      1. \`${OPTION_ELEMENT}\` or \`${OPTION_ATTRIBUTE}\` forces components either to be elements or attributes.
+      1. \`${OPTION_ELEMENT}\` or \`${OPTION_ATTRIBUTE}\` forces components either to be elements or attributes. It can be both (wrapped in an array, see example #4)
```
mgechev: Would you change this to: `forces components to be used as either elements, attributes, or both (not recommend)`.

PapsOu:
@mgechev : I've added your suggestion.

PR #706, opened 2018-08-29
fix(no-input-rename): some failures being reported incorrectly

### LINKED ISSUE #705: no-input-rename fails if input alias is prefixed with directive selector
Currently codelyzer doesn't allow to rename inputs of a directive other than its selector.

```typescript
@Directive({
  selector: '[appHighlight]',
})
export class HighlightDirective {
  // works with codelyzer 4.4
  @Input('appHighlight') visible = true;

  // still an error
  @Input('appHighlightColor') color = 'yellow';
}
```

```html
<h1 appHighlight appHighlightColor="red">Demo</h1>
```

In example above `HighlightDirective#color` have alias prefixed with directive's selector to prevent name collisions within the template(if component or other directive have same input prop). 

If this means anything, `@angular/material` uses this convention(for example: https://github.com/angular/material2/blob/6.4.6/src/lib/sort/sort.ts#L76)

Thank you for this amazing project!

rafaelss95:
Fixes #705.


test/noInputRenameRule.spec.ts
```diff
@@ -111,6 +111,40 @@ describe(ruleName, () => {
           source
         });
       });
+
+      it("should fail when an input alias is prefixed by directive's selector, but the suffix does not match the property name", () => {
```
mgechev: Would you add specs for attribute selectors?

rafaelss95: Hi @mgechev I'm really sry for delay... but right now I can't do this because I'm traveling...

mgechev: No worries

mgechev:
https://github.com/mgechev/codelyzer/pull/723 incorporated this functionality.

PR #723, opened 2018-11-06
fix(no-input-rename): some failures being reported incorrectly

Moeriki:
Update on #705 / #706.

Implement secondary test for successful `tag[attr]` validation, as I think was requested.


mgechev:
I'll look at this tomorrow.

Issue #623, opened 2018-05-11
prefixed with ./. the url of template or style file


wKoza:
Example:

This code

```typescript
@Component({
  selector: 'toh-heroes',
  templateUrl: 'heroes.component.html',
  styleUrls:  ['heroes.component.css']
})
```

should fail.

wKoza:
see [Style guide 05-04](https://angular.io/guide/styleguide#style-05-04)

PR #725, opened 2018-11-23
feat(rule): add new Rule RelativePathExternalResourcesRule

wKoza:
The ./ prefix is standard syntax for relative URLs; don't depend on Angular's current ability to do without that prefix.

Rationale: A component relative URL requires no change when you move the component files, as long as the files stay together.

related to #623

### DIFF
--- a/src/relativeUrlPrefixRule.ts
+++ b/src/relativeUrlPrefixRule.ts
@@ -6,7 +6,7 @@ import * as ts from 'typescript';
 export class Rule extends Rules.AbstractRule {
   static readonly metadata: IRuleMetadata = {
     description: "The ./ prefix is standard syntax for relative URLs; don't depend on Angular's current ability to do without that prefix.",
-    descriptionDetails: 'See more at https://angular.io/guide/styleguide#style-05-04.',
+    descriptionDetails: 'See more at https://angular.io/styleguide#style-05-04.',
     rationale: 'A component relative URL requires no change when you move the component files, as long as the files stay together.',
     ruleName: 'relative-url-prefix',
     options: null,
@@ -51,14 +51,14 @@ export class RelativePathExternalResourcesRuleWalker extends NgWalker {
   }
 
   private checkTemplateUrl(url: string, initializer: ts.StringLiteral) {
-    if (url && !/^.\/./i.test(url)) {
+    if (url && !/^\.\/[^\.\/|\.\.\/]/.test(url)) {
       this.addFailureAtNode(initializer, Rule.FAILURE_STRING);
     }
   }
 
   private checkStyleUrls(token: ts.StringLiteral) {
     if (token && token.text) {
-      if (!/^.\/./i.test(token.text)) {
+      if (!/^\.\/[^\.\/|\.\.\/]/.test(token.text)) {
         this.addFailureAtNode(token, Rule.FAILURE_STRING);
       }
     }
--- a/src/relativeUrlPrefixRule.ts
+++ b/src/relativeUrlPrefixRule.ts
@@ -15,7 +15,7 @@ export class Rule extends Rules.AbstractRule {
     typescriptOnly: true
   };
 
-  static readonly FAILURE_STRING = 'The ./ prefix is standard syntax for relative URLs. (https://angular.io/guide/styleguide#style-05-04)';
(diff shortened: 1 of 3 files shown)


src/relativeUrlPrefixRule.ts
```diff
@@ -0,0 +1,66 @@
...
+    descriptionDetails: 'See more at https://angular.io/guide/styleguide#style-05-04.',
+    rationale: 'A component relative URL requires no change when you move the component files, as long as the files stay together.',
+    ruleName: 'relative-url-prefix',
+    options: null,
+    optionsDescription: 'Not configurable.',
+    type: 'maintainability',
+    typescriptOnly: true
+  };
+
+  static readonly FAILURE_STRING = 'The ./ prefix is standard syntax for relative URLs. (https://angular.io/guide/styleguide#style-05-04)';
```
mgechev: https://angular.io/styleguide#style-05-04 should also work

src/relativeUrlPrefixRule.ts
```diff
@@ -0,0 +1,66 @@
+import { IOptions, IRuleMetadata, RuleFailure, Rules } from 'tslint/lib';
+import { SourceFile } from 'typescript/lib/typescript';
+import { NgWalker } from './angular/ngWalker';
+import * as ts from 'typescript';
+
+export class Rule extends Rules.AbstractRule {
+  static readonly metadata: IRuleMetadata = {
+    description: "The ./ prefix is standard syntax for relative URLs; don't depend on Angular's current ability to do without that prefix.",
+    descriptionDetails: 'See more at https://angular.io/guide/styleguide#style-05-04.',
```
mgechev: We can use https://angular.io/styleguide#style-05-04 instead.

test/relativeUrlPrefixRule.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+          ruleName,
+          message: Rule.FAILURE_STRING,
+          source
+        });
+      });
+
+      it('should fail when a relative URL is prefixed by ../', () => {
+        const source = `
+                    @Component({
+                        templateUrl: '../foobar.html'
```
mgechev: Do we consider `./../foobar.html` correct in this case?

src/relativeUrlPrefixRule.ts
```diff
@@ -0,0 +1,66 @@
...
+    return this.applyWithWalker(new RelativePathExternalResourcesRuleWalker(sourceFile, this.getOptions()));
+  }
+}
+
+export class RelativePathExternalResourcesRuleWalker extends NgWalker {
+  constructor(sourceFile: SourceFile, options: IOptions) {
+    super(sourceFile, options);
+  }
+
+  visitClassDecorator(decorator: ts.Decorator) {
```
mgechev: This method can be simplified if you use `visitNgComponent`.

test/relativeUrlPrefixRule.spec.ts
wKoza: good catch

src/relativeUrlPrefixRule.ts
wKoza: Using `visitNgComponent` was my first thought. Unfortunately, I can't access to 'templateUrl' et 'styleUrls'. It seams more efficient with inline template.

test/relativeUrlPrefixRule.spec.ts
```diff
@@ -105,6 +120,21 @@ describe(ruleName, () => {
           source
         });
       });
+
+        it('should fail when a relative URL is prefixed by ../', () => {
+            const source = `
+                    @Component({
+                        templateUrl: '.././foobar.html'
```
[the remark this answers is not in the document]
wKoza: The style guide Angular says: 
> Why? A component relative URL requires no change when you move the component files, **as long as the files stay together**.

I understood that like 'in the same folder' (mainly in an Angular CLI environment)  but, indeed, it can lead to different interpretations.

Our rationale is `when you move the component files`. So, we should accept all relative urls. 
In my IDE, if I move a component with a relative url, my IDE fixes the new path automatically. 
I agree that `./../` is unclean but, for me, it's the same thing than `../`.

Do you know what I mean ? I can fix it quickly with this in mind (we accept all relative paths).

PR #735, opened 2019-01-30
fix(allow-template-call-expression): allow $any in expressions

mattlewis92:
Hey Minko and co!

`$any` can be used to bypass issues with template typings when `fullTemplateTypeCheck` is on, but codelyzer doesn't seem to know about it, this PR fixes that

I'm not very familiar with how this project works so I've got a few open questions about my implementation:
1. Should the fix be applied in `visitFunctionCall` as well?
2. Should the fix be applied to the parent `RecursiveAngularExpressionVisitor` class instead?
3. Should a new rule be added (in a follow up PR) to disallow the use of `$any`, similar to the [tslint rule](https://palantir.github.io/tslint/rules/no-any/)?

### DIFF
--- a/src/noTemplateCallExpressionRule.ts
+++ b/src/noTemplateCallExpressionRule.ts
@@ -33,6 +33,8 @@ class TemplateVisitor extends BasicTemplateAstVisitor {
 }
 
 class ExpressionVisitor extends RecursiveAngularExpressionVisitor {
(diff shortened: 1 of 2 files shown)


wKoza:
Thanks @mattlewis92 ,

1- visitMethodCall  seems to be the good place for `$any()` ; 
2 - No, your are in the good place also.
3 - You can propose a rule in that

PR #737, opened 2019-02-04
feat(component-change-detection): add change detection strategy rule

### LINKED ISSUE #135: Feature request: Enforce value of @Component.changeDetection
In the project I'm currently working on, it'd be useful to ensure that all components are set to use `changeDetection: ChangeDetectionStrategy.OnPush`.

mattlewis92:
Please let me know if there's any changes I need to make! 😄 

Closes #135

### DIFF
--- a/src/componentChangeDetectionRule.ts
+++ b/src/componentChangeDetectionRule.ts
@@ -1,0 +1,79 @@
+import { Utils } from 'tslint/lib';
+import * as Lint from 'tslint';
+import * as ts from 'typescript';
+import { getDecoratorArgument, getDecoratorName } from './util/utils';
+import { sprintf } from 'sprintf-js';
+import { NgWalker } from './angular';
+
+export class Rule extends Lint.Rules.AbstractRule {
+  public static metadata: Lint.IRuleMetadata = {
+    ruleName: 'component-change-detection',
+    type: 'functionality',
+    description: 'Enforce the preferred component change detection type as ChangeDetectionStrategy.OnPush.',
+    descriptionDetails: Utils.dedent`
+      See more at https://angular.io/api/core/ChangeDetectionStrategy
+    `,
+    options: null,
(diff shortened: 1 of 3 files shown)


src/componentChangeDetectionRule.ts
```diff
@@ -0,0 +1,113 @@
...
+const OPTION_DEFAULT = 'Default';
+const OPTION_ON_PUSH = 'OnPush';
+
+export type ChangeDetectionType = 'Default' | 'OnPush';
+
+export class Rule extends Lint.Rules.AbstractRule {
+  public static metadata: Lint.IRuleMetadata = {
+    ruleName: 'component-change-detection',
+    type: 'functionality',
+    description: 'Enforce the preferred component change detection type (Default or OnPush).',
```
wKoza: I'm not sure it's necessary to have an option.  We want to have a check on use of `OnPush`, not of `Default`

mattlewis92: Thanks so much for the quick review! I was debating about having the option as well, I suppose if angular ever offered a 3rd change detection strategy then we could just add an option for that that default it to OnPush so it wasn't breaking. WDYT?

wKoza: In your implementation, if we use the `DEFAULT` option, we have to set the property `changeDetection: ChangeDetectionStrategy.Default`. It's weird for a default property, no ?

mattlewis92: Yeah agreed, that is kind of weird, I'll just remove the option and keep it as `OnOpen` 👍

mattlewis92:
@wKoza removed the option, LMK what you think! 😄

wKoza:
LGTM ! Thks!

mgechev:
Let's release this once we introduce support for 8.0.0.

PR #743, opened 2019-02-08
feat(rule): only th element can have scope

mohammedzamakhan:


mgechev:
@mohammedzamakhan would you fix the merge conflict?

mohammedzamakhan:
Done 👍

Issue #615, opened 2018-05-06
templates-no-negated-async not reporting some failures


rafaelss95:
**Reproduction:**

```JSON
{
  "rules": {
    "templates-no-negated-async": true
  }
}
```

```ts
@Component({
  selector: 'app-test',
  template: `
    <div *ngFor="let a of collection; trackBy: trackByFn"></div>
    {{ (foo | async) == false }} // not reporting failure here
    {{ !(bar | async) }} // not reporting failure here
  `,
})
export class TestComponent {}

@Component({
  selector: 'app-test',
  template: `
    <div *ngIf="!(a | async)"></div> // idk if it's intentional, but it's not reporting failure
    {{ (foo | async) == false }} // not reporting failure here
    {{ !(bar | async) }} // not reporting failure here
    {{ !(isHandset | async) ? 'dialog' : 'navigation' }} // not reporting failure here
  `,
})
export class TestComponent {}
```

rafaelss95:
@wKoza, can you apply the label `bug` here?

rafaelss95:
I've started to looking at this issue and I found what's the problem:

https://github.com/mgechev/codelyzer/blob/5a84041253dc07ffd6445bef5d6d6733ff43fecd/src/templatesNoNegatedAsyncRule.ts#L28

For the first template above the results are:

`endLeftSpan = 28`
`startRightSpan = 33`
`this.codeWithMap.code.slice(endLeftSpan, startRightSpan);` returns "t a o" (from `*ngFor` line).

It seems that there is a real problem with multiline templates. It's only getting the code from the first line, I really don't know why.

<hr>

## Question:

Why do we always assume that the expression is being compared to false? What if the user put `!` trying to represent a `nil` check (null/undefined)?

```
@Component({
  selector: 'app-test',
  template: '<div *ngIf="!(data$ | async)">Some content</div>'
})
class TestComponent implements OnInit {
  data$: Observable<User | undefined>;

  ngOnInit() {
    this.data$ = someServiceThatCanReturnUserOrUndefined;
  }
}
```

In this case, an user is trying to use `!` to check if `Observable` emits a undefined or even a null value, so the expression **could** be changed to:

```
<div *ngIf="(data$ | async) == null">Some content</div>

<!-- or -->

<div *ngIf="(data$ | async) === undefined">Some content</div>
```

... instead, we replace the expression with `=== false`, what brings an unexpected behavior, a completely unsafe replace.

<hr>

[...]

PR #754, opened 2019-02-10
refactor: general rename/clean up rules

### LINKED ISSUE #653: About rule names
Template rules are a little bit mixed, some start with `template`, others with `templates`, others with nothing.
I want to propose to normalize them with the prefix `template-`:

`banana-in-box` -> `template-banana-in-box`;
`i18n` -> `template-i18n`;
`templates-no-negated-async` -> `template-no-negated-async`
`trackBy-function` -> `template-use-trackBy-function` (not sure about the capitalized "B")
`angular-whitespace` -> `template-whitespace`

<hr>

Also, another suggestions for names would be:

All the `*-life-cycle-*` rules should be renamed to `*-lifecycle-*`, as mentioned in some issues/PRs;
`no-output-named-after-standard-event` should be renamed to `no-output-standard-event` or something like that since this rule cares precisely about exact output names.

rafaelss95:
| Current name                            |        Proposed name                  |
| :------------:                          |        :------------:                 |
| contextual-life-cycle                   |     contextual-lifecycle              |
| no-conflicting-life-cycle-hooks         |     no-conflicting-lifecycle          |
| no-life-cycle-call                      |     no-lifecycle-call                 |
| use-life-cycle-interface                |     use-lifecycle-interface           |
| decorator-not-allowed                   |     contextual-decorator              |
| enforce-component-selector              |     use-component-selector            |
| no-output-named-after-standard-event    |     no-output-native                  |
| use-host-property-decorator             |     no-host-metadata-property         |
| use-input-property-decorator            |     no-inputs-metadata-property       |
| use-output-property-decorator           |     no-outputs-metadata-property      |
| no-queries-parameter                    |     no-queries-metadata-property      |
| pipe-impure                             |     no-pipe-impure                    |
| use-view-encapsulation                  |     use-component-view-encapsulation  |
| i18n                                    |     template-i18n                     |
| banana-in-box                           |     template-banana-in-box            |
| no-template-call-expression             |     template-no-call-expression       |
| templates-no-negated-async              |     template-no-negated-async         |
| trackBy-function                        |     template-use-track-by-function    |

<hr>

## Points:

**1) EnforceComponentSelector -> UseComponentSelector**: I decided to rename it to make it sync with other "use" rules... let me know what do you think about it.
**2) NoOutputNamedAfterStandardEventRule -> NoOutputNative**: not sure if it's a good name, please don't hesitate to suggest one :)
~~**3) TemplatesNoNegatedAsync -> TemplateNoNegatedAsync**: I'm not sure renaming it is a good idea, because in addition to this rule returning some false positives, it's not useful at all (see https://github.com/mgechev/codelyzer/issues/615#issuecomment-405132209)~~ **resolved in #615**.
**4)** Since we are renaming, what about those below?

`pipe-impure` -> `no-pipe-impure`
`use-host-property-decorator` -> `no-host-parameter` (to be sync with `no-queries-parameter`)
`use-input-property-decorator` -> `no-inputs-parameter` (to be sync with `no-queries-parameter`)
`use-output-property-decorator` -> `no-outputs-parameter` (to be sync with `no-queries-parameter`)

Closes #653.


mgechev:
This will require schematics transform of `tslint.json` when users execute `ng update`.

We'll have to discuss the change with @alexeagle and @hansl.

README.md
```diff
@@ -96,38 +96,38 @@ Create the following `tslint.json` file like:
...
+    "template-conditional-complexity": [true, 4],
     "template-cyclomatic-complexity": [true, 5],
-    "template-conditional-complexity": [true, 4]
+    "template-i18n": [true, "check-id", "check-text"],
+    "template-no-negated-async": true,
+    "template-use-track-by-function": true,
+    "use-component-selector": true,
+    "use-host-property-decorator": true,
+    "use-input-property-decorator": true,
+    "use-lifecycle-interface": true,
```
mgechev: We use this internally. Need to make sure we update g3.

src/contextualLifecycleRule.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.validateDecorator(controller, decorator, METADATA_TYPE_LIFECYCLE_MAPPER.Injectable);
+    super.visitNgInjectable(controller, decorator);
+  }
+
+  protected visitNgPipe(controller: ClassDeclaration, decorator: Decorator): void {
+    this.validateDecorator(controller, decorator, METADATA_TYPE_LIFECYCLE_MAPPER.Pipe);
+    super.visitNgPipe(controller, decorator);
+  }
+
+  private validateDecorator(
```
mgechev: Good job with this refactoring!

mgechev:
The renames you suggest make sense.

src/noConflictingLifecycleHooksRule.ts
```diff
@@ -0,0 +1,100 @@
...
+  visitClassDeclaration(node: ClassDeclaration): void {
+    this.validateInterfaces(node);
+    this.validateMethods(node);
+    super.visitClassDeclaration(node);
+  }
+
+  private validateInterfaces(node: ClassDeclaration): void {
+    const declaredLifecycleInterfaces = getDeclaredLifecycleInterfaces(node);
+
+    for (const key of Object.keys(LIFECYCLE_INTERFACE_CONFLICT_MAPPER)) {
```
mgechev: Let's drop the for loop here, we don't really need it, nor `LIFECYCLE_INTERFACE_CONFLICT_MAPPER`

src/noConflictingLifecycleHooksRule.ts
```diff
@@ -0,0 +1,100 @@
...
+      );
+
+      this.addFailureAtNode(node, failure);
+    }
+  }
+
+  private validateMethods(node: ClassDeclaration): void {
+    const declaredLifecycleMethods = getDeclaredLifecycleMethods(node);
+
+    for (const key of Object.keys(LIFECYCLE_METHOD_CONFLICT_MAPPER)) {
```
mgechev: Same here.

src/noLifecycleCallRule.ts
```diff
@@ -0,0 +1,47 @@
...
+  };
+
+  static readonly FAILURE_STRING = 'Avoid explicit calls to lifecycle hooks';
+
+  apply(sourceFile: SourceFile): RuleFailure[] {
+    return this.applyWithWalker(new ExpressionCallMetadataWalker(sourceFile, this.getOptions()));
+  }
+}
+
+export class ExpressionCallMetadataWalker extends NgWalker {
```
mgechev: Do we need to export this?

src/templateBananaInBoxRule.ts
```diff
@@ -0,0 +1,65 @@
+import { BoundEventAst } from '@angular/compiler';
+import { IRuleMetadata, Replacement, RuleFailure } from 'tslint';
+import { AbstractRule } from 'tslint/lib/rules';
+import { SourceFile } from 'typescript';
+import { NgWalker } from './angular/ngWalker';
+import { BasicTemplateAstVisitor } from './angular/templates/basicTemplateAstVisitor';
+
+const INVALID_BOX = /\[(.*)\]/;
```
mgechev: `/\[(.*)\]/.test('[(ngModel)]') && /\[(.*)\]/.test('([ngModel])')` which seems incorrect.

src/noLifecycleCallRule.ts
rafaelss95: Not necessarily, however, for some reason, we export the walker's in almost all rules.

src/templateNoNegatedAsyncRule.ts
```diff
@@ -22,57 +46,31 @@ class TemplateToNgTemplateVisitor extends RecursiveAngularExpressionVisitor {
...
+    } = ast;
+
+    this.addFailureFromStartToEnd(spanStart, spanEnd, errorMessage);
+  }
+
   private validateBinary(ast: Binary): void {
     const { left, operation, right } = ast;
 
-    if (!isAsyncBinding(left) || !(right instanceof LiteralPrimitive) || right.value !== false || operation !== unstrictEqualityOperator) {
+    if (!isAsyncBinding(left) || !(right instanceof LiteralPrimitive) || right.value !== false || operation !== unstrictEqualityOperator)
```
mgechev: Let's return the braces

src/useComponentSelectorRule.ts
```diff
@@ -0,0 +1,41 @@
...
+export class UseComponentSelectorValidatorWalker extends NgWalker {
+  protected visitNgComponent(metadata: ComponentMetadata): void {
+    this.validateComponent(metadata);
+    super.visitNgComponent(metadata);
+  }
+
+  private validateComponent(metadata: ComponentMetadata): void {
+    if (metadata.selector) return;
+
+    const failure = getFailureMessage(metadata.controller.name!.text);
```
mgechev: We're very confident here. Do we know for sure that `name` is not going to be `null`?

test/contextualLifecycleRule.spec.ts
```diff
@@ -0,0 +1,113 @@
+import { getFailureMessage, Rule } from '../src/contextualLifecycleRule';
```
mgechev: I'd prefer to keep the previous specs. This is not very readable, it's hard to keep track what are the interfaces and the methods.

test/noLifecycleCallRule.spec.ts
```diff
@@ -1,51 +1,46 @@
-import { lifecycleHooksMethods, LifecycleHooksMethods, Rule } from '../src/noLifeCycleCallRule';
+import { Rule } from '../src/noLifecycleCallRule';
```
mgechev: Same here. I see we've merged the template string snippets, but I have no idea what we are testing like this. Would you mind updating? Thanks 🙇🏼‍♂️

src/templateBananaInBoxRule.ts
rafaelss95: You're right, however, the previous regex (which one is different only by a question mark) was also wrong, it gave the same result for the tests you wrote above.

<hr>

After investigating why the specs pass even with this wrong regex, I discovered that the ast's name just cuts the parens, so:

**Test 1**: `<input type="text" name="foo" ([ngModel])="foo">`, test with regex returns true (the ast's name is just `[ngModel]`);
**Test 2**: `<div (emitter)="emitter"></div>` -> test with regex returns false (ast's name is `emitter`);
**Test 3**:  `<input type="text" name="foo" [ngModel]="foo">` -> test with regex returns false (ast's name is `ngModelChange`);
**Test 4**: `<input type="text" name="foo" [(ngModel)]="foo">` -> never call `visitEvent` method.

<hr>

I changed the regex a bit to: `/^\[(?!\()(.*)(?<!\))\]$/`.

src/noConflictingLifecycleHooksRule.ts
rafaelss95: Done.

src/noConflictingLifecycleHooksRule.ts
rafaelss95: Done.

src/useComponentSelectorRule.ts
rafaelss95: Hmm, to be sincere, I really don't know if we can receive a `null` name here. It was copied from `enforceComponentSelector`, anyway, I'll add a check for this :)

test/noLifecycleCallRule.spec.ts
rafaelss95: No problem, I'll update ASAP.

test/contextualLifecycleRule.spec.ts
rafaelss95: No problem, I'll update ASAP.

src/templateBananaInBoxRule.ts
mgechev: Yes, previous regex was absolutely wrong, I noticed that. Thanks for the update!

src/noLifecycleCallRule.ts
mgechev: 👍

rafaelss95:
Haha, thanks :) Btw, it's ready for the next round when you have time 💥

src/templateBananaInBoxRule.ts
wKoza: yes @rafaelss95 , this rule used `visitEvent`, what explains this behavior.

test/useLifecycleInterfaceRule.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+          }
+        `;
+        assertSuccess(ruleName, source);
+      });
+    });
+
+    describe('valid declaration of lifecycle hooks, using ng.hookName', () => {
+      it('should succeed when lifecycle hook is used with its interface', () => {
+        const source = `
+          class ${className} implements ng.OnInit {
```
wKoza: Question: using namespace is a good practice ?

rafaelss95: I really don't know, but I never saw someone using this in real world.

src/contextualLifecycleRule.ts
```diff
@@ -0,0 +1,64 @@
...
+    allowedHookMethods: ReadonlySet<LifecycleMethodKeys>
+  ): void {
+    if (!controller.name) return;
+
+    const decoratorName = decorator.getText();
+
+    for (const member of controller.members) {
+      const { name: memberName } = member;
+
+      if (!memberName) continue;
```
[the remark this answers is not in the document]
rafaelss95: I've never seen a case where this could be null! I did it to jump the `!` 😛

src/contextualLifecycleRule.ts
```diff
@@ -0,0 +1,64 @@
...
+    this.validateDecorator(controller, decorator, METADATA_TYPE_LIFECYCLE_MAPPER.Pipe);
+    super.visitNgPipe(controller, decorator);
+  }
+
+  private validateDecorator(
+    controller: ClassDeclaration,
+    decorator: Decorator,
+    allowedHookMethods: ReadonlySet<LifecycleMethodKeys>
+  ): void {
+    if (!controller.name) return;
```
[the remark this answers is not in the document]
rafaelss95: I've never seen a case where this could be null! I did it to jump the `!` 😛

test/useLifecycleInterfaceRule.spec.ts
wKoza: @mgechev , what do you think about it ?

mgechev: This may have some implications in the dead code elimination, but it's fine to exist in our test cases.

mgechev:
@rafaelss95 would you resolve the merge conflict? I think we're ready to merge.

rafaelss95:
@mgechev Sure! What about my three questions?

src/contextualLifecycleRule.ts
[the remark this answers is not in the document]
mgechev:
```ts
@Component({...})
export default class { ... }
```

Seems valid to me. In this case `name` is undefined.

src/contextualLifecycleRule.ts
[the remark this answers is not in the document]
mgechev: Let's keep it.

mgechev:
LGTM, the extra checks seem fine.

rafaelss95:
@mgechev I was talking about the questions in the PR description (below the table) :)

mgechev:
Oh, I was sure that I responded to these questions but it looks I didn't. Yes, it makes sense to rename the rest of the rules as well.

src/contextualLifecycleRule.ts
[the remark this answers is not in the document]
wKoza: hmm, In the Angular world, I'm not sure we can omit a Type.  But, @mgechev , your example is interesting:
If we create a simple class  like that:
```typescript
export default class { ... }
```

and we launch `ng lint`, we have a big big crash [here] (https://github.com/mgechev/codelyzer/blob/master/src/propertyDecoratorBase.ts#L41) and many other places ! 

@mgechev , @rafaelss95 , I'm not very fan of the `!`  Non-Null Assertion. We cannot say that 
```typescript
const className = metadata.controller.name!.text
```
is the same that
```typescript
if (!controller.name) return;
 const className = controllerName.text;
```
If we accept `metadata.controller.name!.text`, we should not write `if (!controller.name) return;` 

Hence my 'innoncent' question.

So, we must keep this test, but we must also use this test everywhere and stop to use the Non-Null Assertion.

Are you agree ?

rafaelss95:
Hey @mgechev I've updated the code and the table in OP with the new proposed names.

<hr>

I took a look in (almost) all the rules today and found some more rule names that can be improved:

`component-change-detection`:
  - Suggestions:
    - no-component-default-change-detection
    - prefer-component-change-detection-on-push
    - use-component-change-detection-on-push

`relative-url-prefix`:
  - Suggestions:
    - prefer-component-relative-url-prefix
    - use-component-relative-url-prefix

`no-attribute-parameter-decorator`:
  - Suggestions:
    - no-attribute-decorator
    - prefer-input-over-attribute-decorator

Let me know your opinion.

mgechev:
@rafaelss95 let's rename the other rules as part of another PR:

* prefer-on-push-component-change-detection
* no-attribute-decorator

Sound good. About the relative prefix, it's still not obvious what's about. Maybe we should be more explicit about it.

rafaelss95:
>... About the relative prefix, it's still not obvious what's about. Maybe we should be more explicit about it.

Do you have any suggestion?

rafaelss95:
>... About the relative prefix, it's still not obvious what's about. Maybe we should be more explicit about it.

@mgechev Do you have any suggestion?

mgechev:
Let's not updated it for now.

Issue #633, opened 2018-05-26
Feature request: template-no-any


rafaelss95:
A rule to disallow the `$any(...)` usage:

```ts
interface Human {
  name: string;
}

@Component({
  // $any() nullifies the compile-time benefits of the type system.
  template: '<p>Hello {{ $any(human).age }}</p>'
})
export class AppComponent {
  human: Human = { name: 'Human' };
  // human has no field age, so the template should not compile
}
```

Similar to https://palantir.github.io/tslint/rules/no-any/

PR #755, opened 2019-02-11
feat: add template-no-any rule

rafaelss95:
Please, check if the description, rationale and failure are clear enough :)

Issue: #633


src/index.ts
```diff
@@ -29,6 +29,7 @@ export { Rule as PreferOutputReadonlyRule } from './preferOutputReadonlyRule';
 export { Rule as TemplateConditionalComplexityRule } from './templateConditionalComplexityRule';
 export { Rule as TemplateCyclomaticComplexityRule } from './templateCyclomaticComplexityRule';
 export { Rule as TemplateAccessibilityTabindexNoPositiveRule } from './templateAccessibilityTabindexNoPositiveRule';
+export { Rule as TemplateNoAnyRule } from './templateNoAnyRule';
```
mgechev: Would you also add the rule to the readme?

rafaelss95:
> Excellent! Would you only add this rule to the readme?

Isn't it better to add it only when we publish an major version?

wKoza:
@rafaelss95 , maybe, can you change the Readme of [this branch](https://github.com/mgechev/codelyzer/pull/756/files) ?

PR #760, opened 2019-02-11
feat(rule): no distracting elements should be used

mohammedzamakhan:


src/templateNoDistractingElementsRule.ts
```diff
@@ -0,0 +1,51 @@
...
+    description: 'Enforces that no distracting elements are used',
+    options: null,
+    optionsDescription: 'Not configurable.',
+    rationale: 'Elements that can be visually distracting can cause accessibility issues with visually impaired users.',
+    ruleName: 'template-no-distracting-elements',
+    type: 'functionality',
+    typescriptOnly: true
+  };
+
+  static readonly FAILURE_STRING = 'Avoid using <%s> elements as they create visual accessibility issues.';
```
mgechev: `<%s>` -> `<%s/>`

src/templateNoDistractingElementsRule.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+class TemplateNoDistractingElementsVisitor extends BasicTemplateAstVisitor {
+  visitElement(prop: ElementAst, context: any): any {
+    this.validateElement(prop);
+    super.visitElement(prop, context);
+  }
+
+  private validateElement(el: ElementAst): void {
+    if (el.name === 'marquee' || el.name === 'blink') {
```
wKoza: When I see [this link](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/blink#Browser_compatibility), I'm not sure that we have to test 'blink'.

mohammedzamakhan: I think its fine, since all the accessibility checkers and linters check for it. Sometimes developers use code generated or from old references.

PR #761, opened 2019-02-12
feat(rule): click event should be accompanied with key event

mohammedzamakhan:


src/templateClickEventsHaveKeyEventsRule.ts
```diff
@@ -0,0 +1,55 @@
+import { ElementAst } from '@angular/compiler';
+import { IRuleMetadata, RuleFailure, Rules } from 'tslint/lib';
+import { SourceFile } from 'typescript/lib/typescript';
+import { NgWalker } from './angular/ngWalker';
+import { BasicTemplateAstVisitor } from './angular/templates/basicTemplateAstVisitor';
+
+export class Rule extends Rules.AbstractRule {
+  static readonly metadata: IRuleMetadata = {
+    description: 'Ensures that the Click Event is accompanied with at least one Key Events keyup, keydown or keypress',
```
mgechev: click event could be lower case

mgechev: Same for key events

PR #772, opened 2019-02-22
fix(rule):  don't check keyup events for some elements

### LINKED ISSUE #771: bug: template-click-events-have-key-events: Exclude interactive elements
Make sure that interactive elements like button are excluded

mohammedzamakhan:
fixes #771


src/templateClickEventsHaveKeyEventsRule.ts
```diff
@@ -37,6 +43,21 @@ class TemplateClickEventsHaveKeyEventsVisitor extends BasicTemplateAstVisitor {
     if (!hasClick) {
       return;
     }
+
+    if (domElements.indexOf(el.name) === -1) {
```
mgechev: How many elements do we have in `domElements`? This linear check might be too slow. If we keep it as a set, the check would be `O(1)`.

src/util/attributesComparator.ts
```diff
@@ -0,0 +1,26 @@
+import { ElementAst, BoundElementProperty, AttrAst } from '@angular/compiler';
+import { getAttributeValue } from './getAttributeValue';
+import { getLiteralValue } from './getLiteralValue';
+
+function attributesComparator(baseAttributes: any = [], el: ElementAst): boolean {
+  const attributes: Array<AttrAst | BoundElementProperty> = [].concat(...(el.attrs as any), ...(el.inputs as any));
```
mgechev: I'd suggest to refactor it to use sets/maps, so we can drop the quadratic check to a linear one.

src/util/isInteractiveElement.ts
```diff
@@ -0,0 +1,100 @@
+import { dom, elementRoles, roles } from 'aria-query';
+
+import { AXObjects, elementAXObjects } from 'axobject-query';
+
+import attributesComparator from './attributesComparator';
+import { ElementAst } from '@angular/compiler';
+
+const domKeys = [...(<string[]>Array.from(dom.keys()))];
```
mgechev: `(<string[]>Array.from(dom.keys()))`, this should be enough. No need to spread the array into another one

src/util/isInteractiveElement.ts
```diff
@@ -0,0 +1,100 @@
...
+const roleKeys: any = [...(<string[]>Array.from(roles.keys()))];
+const elementRoleEntries = [...Array.from(elementRoles)];
+
+const nonInteractiveRoles = new Set(
+  roleKeys.filter(name => {
+    const role = roles.get(name);
+    return !role.abstract && !role.superClass.some(classes => classes.indexOf('widget') !== 0);
```
mgechev: Maybe `every` would be better in this case

src/util/isInteractiveElement.ts
```diff
@@ -0,0 +1,100 @@
...
+      'toolbar'
+    )
+    .filter(name => {
+      const role = roles.get(name);
+      return !role.abstract && role.superClass.some(classes => classes.indexOf('widget') !== 0);
+    })
+);
+
+const nonInteractiveElementRoleSchemas = elementRoleEntries.reduce((accumulator: any, [elementSchema, roleSet]: any) => {
+  if ([...Array.from(roleSet)].every((role): boolean => nonInteractiveRoles.has(role))) {
```
mgechev: Same with the spread here.

src/templateClickEventsHaveKeyEventsRule.ts
mohammedzamakhan: 129 elements

src/util/getAttributeValue.ts
```diff
@@ -7,6 +8,6 @@ export const getAttributeValue = (element: ElementAst, property: string) => {
     return attr.value;
   }
   if (input) {
-    return (<any>input.value).ast.value;
+    return (<any>input.value).ast.value || PROPERTY;
```
rafaelss95: Let's try to avoid these `any` casts.

You can do something like this instead:

```
if (!input || !(input.value instanceof ASTWithSource) || !(input.value.ast instanceof LiteralPrimitive)) {
  return null/undefined;
}

return input.value.ast.value;
```

src/util/attributesComparator.ts
rafaelss95: Also, if possible, can we remove these `any` casts?

src/util/isInteractiveElement.ts
```diff
@@ -0,0 +1,97 @@
...
+  })
+);
+
+const interactiveRoles: any = new Set(
+  []
+    .concat(
+      roleKeys,
+      // 'toolbar' does not descend from widget, but it does support
+      // aria-activedescendant, thus in practice we treat it as a widget.
+      // @ts-ignore
```
rafaelss95: Could you explain why this is necessary?

src/util/attributesComparator.ts
mohammedzamakhan: @rafaelss95 done

@mgechev I have two different arrays, I don't think I can escape the quadratic check here

PR #780, opened 2019-02-27
refactor: improve NgWalker and remove class name from error message

### LINKED ISSUE #778: Codelyzer doesn't support export default
I reedit [my comment](https://github.com/mgechev/codelyzer/pull/754#discussion_r258851555) form the PR #754.

If we create a simple class like that:

```ts
export default class { ... }
```

and we launch `ng lint`, we have a big big crash [here] (https://github.com/mgechev/codelyzer/blob/master/src/propertyDecoratorBase.ts#L41) and many other places !

wKoza:
- improve NgWalker 
- Remove class name from error message
- Prevent an error when a class has no name.

closes #778

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -582,6 +582,22 @@
         }
       }
     },
+    "aria-query": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/aria-query/-/aria-query-3.0.0.tgz",
+      "integrity": "sha1-ZbP8wcoRVajJrmTW7uKX8V1RM8w=",
+      "requires": {
+        "ast-types-flow": "0.0.7",
+        "commander": "^2.11.0"
+      },
+      "dependencies": {
+        "commander": {
+          "version": "2.19.0",
+          "resolved": "https://registry.npmjs.org/commander/-/commander-2.19.0.tgz",
+          "integrity": "sha512-6tvAOO+D6OENvRAh524Dh9jcfKTYDQAqvqezbCW82xj5X0pSrcpxtvRKHLG0yBY6SD7PSDrJaj+0AiOcKVd1Xg=="
+        }
+      }
+    },
     "arr-flatten": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/arr-flatten/-/arr-flatten-1.1.0.tgz",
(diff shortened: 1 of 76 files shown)


rafaelss95:
Nice refactor, @wKoza!

What about also removing attribute names from the errors (for the reason I exposed in #778)?

Ex. (`noOutputNativeRule`):

**Current message:** "The output property "%s" should not be named or renamed as a native event"
**Proposed message:** "Output properties should not be named or renamed as a native event" (... or something like this, I think you got the idea)

wKoza:
yep, I know but we will lost some informations. But, I can do the exercice:

> The name of a component should end with the suffix %s (https://angular.io/styleguide#style-02-03
The name of a component should end with the predefined suffix (https://angular.io/styleguide#style-02-03
  
> The selector of a component should have prefix "%s" (https://angular.io/styleguide#style-02-07)
The selector of a component should have the predefined prefix (https://angular.io/styleguide#style-02-07)

> The selector of a component should have one of the prefixes "%s" (https://angular.io/styleguide#style-02-07)
The selector of a component should have one of the predefined prefixes (https://angular.io/styleguide#style-02-07)

> The selector of a component should be named %s (https://angular.io/styleguide#style-05-02)
The selector of a component should be named correctly (https://angular.io/styleguide#style-05-02)

> The selector of a component should be used as %s (https://angular.io/styleguide#style-05-03)
The selector of a component should be used as preset (https://angular.io/styleguide#style-05-03)

> 'The decorator "%s" is not allowed for a class decorated with "%s"'
'This decorator is not allowed for a class decorated like this'

...

I think that is a pity, The messages become mysterious.

mgechev:
Any ideas how to improve the error messages?

wKoza:
We can definitely change few error messages but I  I think it is a shame to lost some informations. Otherwise, we can add a code as suffix like 

> CL001: 'The decorator "%s" is not allowed for a class decorated with "%s

if we want to help users who want to make statistics.

mgechev:
@wKoza yes it makes sense. I'd not want to lose information either.

rafaelss95:
@mgechev @wKoza maybe you misunderstood what I said. I did not say to remove **all** the names. I tried to say that we could remove *redundant* attribute names for _some_ messages, as in the example I cited above and copied below.

Some of them are important. In other words, we must investigate each case separately.

> Ex. (`noOutputNativeRule`):
> 
> **Current message:** "The output property "%s" should not be named or renamed as a native event"
> **Proposed message:** "Output properties should not be named or renamed as a native event" (... or something like this, I think you got the idea)

**Reason**: The failure comes with the exact line that fails, where the user has the property name.

wKoza:
I've changed few error messages when it was no pertinent + rebase

rafaelss95:
@wKoza 

I guess it's unmaintainable add a spec to check if default classes "break the linter" for each rule. Can't we add it in a single place?

Btw, I've wrote some comments/suggestions... can you rebase it, so I can send my review updated? Also... if possible, can you separate into small commits? It's kind hard to review this :/

wKoza:
When I see the number of conflict, I prefer close this PR. I don't want to block a new release for that.

rafaelss95:
IMHO we shouldn't close this PR. It's an important change and I think we can get everything done before the next major version. The review could be faster if you can separate into small commits...

PR #781, opened 2019-03-02
refactor(rule): rename/clean up 'no-attribute-parameter-decorator'

rafaelss95:
Related to #754.


src/noAttributeDecoratorRule.ts
```diff
@@ -0,0 +1,61 @@
...
+    description: 'Disallows usage of @Attribute decorator.',
+    options: null,
+    optionsDescription: 'Not configurable.',
+    rationale: '@Attribute is considered bad practice. Use @Input instead.',
+    ruleName: 'no-attribute-decorator',
+    type: 'functionality',
+    typescriptOnly: true
+  };
+
+  static readonly FAILURE_STRING = '@Attribute is considered bad practice. Use @Input instead.';
```
rafaelss95: Not sure about this message...

rafaelss95:
Ping @mgechev

src/noAttributeDecoratorRule.ts
rafaelss95: If there's any place in the Angular Style Guide that says it's a bad practice to use `@Attribute` we should link it here, otherwise better rephrase this (and also the `rationale`).

mgechev: LGTM

Issue #777, opened 2019-02-24
About RuleWalker


rafaelss95:
In the [latest TSLint's release](https://github.com/palantir/tslint/blob/master/CHANGELOG.md#v5130), they marked `RuleWalker` as deprecated. Apparently, [RuleWalker-based rules are slow and should be avoided](https://github.com/palantir/tslint/issues/2522). Considering this, what should we do with the current rules? @mgechev @wKoza

wKoza:
@rafaelss95 , `RuleWalker` is deprecated ? Tslint will be also deprecated by the end of the year. There is also a big problem with the new compiler Ivy. There are many impacts.
That's a lot of change in one year. :stuck_out_tongue_winking_eye:

mgechev:
We can catch up on hangouts to discuss what we're going to do with TSLint. I'll ping you over email.

PR #784, opened 2019-03-04
refactor: bump packages versions

rafaelss95:
Currently we maintain 2 `lock` files, but once you try to execute `yarn` with `yarn.lock` and `package-lock.json`, you get the following message:

> warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.

To fix this, I've decided to remove the `package-lock.json`. If you prefer to use `npm`, we can remove `yarn.lock` instead.

<hr>

### Summary:

* **Removed unused packages:**
  - `@types/less`;
  - `minimalist`;

* **Packages that can't be updated because break the rules/tests:**
  - `chai-spies`;
  - `source-map`;

* Move git hooks from package.json to `.huskyrc.json` file.

* Since TSLint is also being upgraded, some rules, which extends `RuleWalker` **directly**, were reworked to use `WalkContext` instead (#777).

* `NgWalker` and `SourceMappingVisitor` `extends` the `RuleWalker` and for now the `deprecation` rule for the `extends` line was just disabled.

* Some Typescript errors are caught by the newer version and fixed properly.

* The tests are green :white_check_mark:


mgechev:
This will take a while to review. It'll be very helpful if you can separate it into multiple PRs or at least multiple commits so I can review them in a sequence?

rafaelss95:
Done, @mgechev.

mgechev:
Wow, that was fast! :)

src/noInputRenameRule.ts
```diff
@@ -90,17 +92,15 @@ export class InputMetadataWalker extends NgWalker {
       (propertyAlias !== propertyName &&
         this.directiveSelectors &&
         this.directiveSelectors.some(x => new RegExp(`^${x}((${toTitleCase(propertyName)}$)|(?=$))`).test(propertyAlias))) ||
-      (whiteListAliases.has(propertyAlias) && propertyName === kebabToCamelCase(propertyAlias))
+      (whiteListAliases.has(propertyAlias) && propertyName === camelize(propertyAlias))
```
mgechev: `kebabToCamelCase` is more semantically correct. We will not be able to `camelize` an arbitrary string. Would you mind reverting this?

mgechev:
@rafaelss95 thanks for updating the AST traversal from visitors to functions!

src/noInputRenameRule.ts
rafaelss95: No problem. I did it because I saw that TSLint has already a function to camelize strings.

PR #787, opened 2019-03-08
fix(rule): 'prefer-inline-decorator' limiting the number of options

rafaelss95:
…assed as options

Currently, the rule allows only eight options/decorators to be passed.


rafaelss95:
@mgechev @wKoza 

Look at this spec:

https://github.com/mgechev/codelyzer/blob/049be667a6c91940b5d203d7e7e09af4ef50b550/test/preferInlineDecoratorRule.spec.ts#L31-L65

<hr>

When I implemented this rule, I had to use `assertFailures` because when I tried to use `assertMultipleAnnotated`, I got `no failures`. Do you know why?

... Btw, the same happens with `template-no-call-expression`: if you test it using `non null assert`, for example, like this:

`{{ obj!.nested1() }}`

... you'll see the failure being reported (in the rule), but in the `spec`, the error isn't being caught.

Any thoughts?

mgechev:
@rafaelss95 not sure where this may come from, I haven't hit this issue but sounds like a problem in `assertMultipleAnnotated`

PR #790, opened 2019-03-17
fix(rule): 'NgModule' not being handled properly for contextual-lifecycle and contextual-decorators

wKoza:


test/contextualLifecycleRule.spec.ts
```diff
@@ -150,6 +150,170 @@ describe(ruleName, () => {
       });
     });
 
+    describe('failure', () => {
```
rafaelss95:
```suggestion

```

We already have a _failure_ **describe** block. You can put the `NgModule` block into it.

rafaelss95:
```suggestion

```

We already have a _failure_ **describe** block. You can put the `NgModule` block into it.

rafaelss95:
Nice catch! ✅ 

I guess this change also fix the `contextual-decorator`, right? If so, let's rephrase the commit.

wKoza:
Actually, we kill two birds with one stone.

rafaelss95:
LGMT!

PR #793, opened 2019-03-19
refactor: walkers/visitors being exported

rafaelss95:
In addition to removing the export from visitors/walkers that contained it, it renames them to a _generic_ name to avoid confusion (ex: for `templateNoAutofocus` rule we are naming its visitor as `TemplateConditionalComplexityVisitor`: https://github.com/mgechev/codelyzer/blob/799382fd1b16bbff9bfb1587b643c4125df8f393/src/templateNoAutofocusRule.ts#L29)... and to keep them consistent.


rafaelss95:
Friendly bump @mgechev :)

PR #794, opened 2019-03-19
refactor(rule): rework prefer-inline-decorator to accept options

rafaelss95:
- The rule `prefer-inline-decorator`:
  - Doesn't report failures for non **call expression** decorators (Actually, the culpirit is the utility function `getDecoratorName`, which throws an error if you try to get a name of a `decorator` that isn't a **call expression**);
  - Doesn't report failures for getters, setters, parameters and parameters properties. To handle this (separately),  options are being added. Users can disable according to their project standards.
  - Its fix is **unsafe**. If the `fix` is applied to a property/method with multiple decorators, it just keeps the first decorator (from top) and aligns the property/method with this remaining decorator. I've opted to remove the `hasFix` _option_, because it would be a pain to handle all possible cases correctly.

### DIFF
--- a/src/preferInlineDecoratorRule.ts
+++ b/src/preferInlineDecoratorRule.ts
@@ -1,81 +1,286 @@
-import { IOptions, IRuleMetadata, Replacement, RuleFailure } from 'tslint/lib';
+import { IRuleMetadata, RuleFailure, WalkContext } from 'tslint/lib';
 import { AbstractRule } from 'tslint/lib/rules';
-import { Decorator, isPropertyDeclaration, SourceFile } from 'typescript';
-import { NgWalker } from './angular/ngWalker';
+import { dedent } from 'tslint/lib/utils';
+import {
+  createNodeArray,
+  Decorator,
+  forEachChild,
+  GetAccessorDeclaration,
+  isGetAccessorDeclaration,
+  isMethodDeclaration,
+  isParameter,
+  isParameterPropertyDeclaration,
+  isPropertyDeclaration,
+  isSetAccessorDeclaration,
+  MethodDeclaration,
+  Node,
+  ParameterDeclaration,
+  ParameterPropertyDeclaration,
+  PropertyDeclaration,
+  SetAccessorDeclaration,
+  SourceFile
+} from 'typescript';
+import { isNotNullOrUndefined } from './util/is-not-null-or-undefined';
+import { objectKeys } from './util/object-keys';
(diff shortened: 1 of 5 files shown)


mgechev:
This will require more time for review. I'll try to push it sometime next week.

src/preferInlineDecoratorRule.ts
```diff
@@ -1,79 +1,299 @@
...
+
+    return this.applyWithFunction(sourceFile, walk, options);
   }
 
   isEnabled(): boolean {
-    return super.isEnabled() && this.ruleArguments.every(ruleArgument => !!(typeof ruleArgument === 'string' && ruleArgument.trim()));
+    return super.isEnabled() && this.areOptionsValid();
+  }
+
+  private areOptionsValid(): boolean {
```
mgechev: If the options are not valid, we should probably throw an error. Wondering why we're unable to validate directly with the schema? Doesn't tslint do this already?

src/preferInlineDecoratorRule.ts
```diff
@@ -1,79 +1,299 @@
...
+};
 
-    const isDecoratorBlacklisted = this.blacklistedDecorators.has(decoratorName);
+const getNonIgnoredDecoratorNames = (
+  walkContext: WalkContext<OptionDictionary>,
+  decorators: ReadonlyArray<Decorator>,
+  optionKey: OptionKeys
+): ReadonlyArray<string> => {
+  return decorators
+    .map(decorator => getDecoratorName(decorator))
```
mgechev: Just `getDecoratorName`

rafaelss95: Good catch!

src/preferInlineDecoratorRule.ts
rafaelss95: So... this is something that I was about to propose to you: use a lib (something like https://github.com/epoberezkin/ajv) to validate JSON schemas.

AFAIK, there's no in-built validation for tslint. All the checks are "hard-coded" in the `isEnabled` method from https://github.com/palantir/tslint/blob/4b8bbf66b7fe7f49122debdeb02164e045b3a691/src/language/rule/abstractRule.ts#L49

rafaelss95: What's your opinion about it 👆 ? @wKoza @mgechev

mgechev: JSON schema validation would be great. We do this in the CLI for builders' arguments.

PR #798, opened 2019-03-24
 fix(rule): 'contextual-decorator' - decorators with arguments, accessors and some missing decorators not being handled

### LINKED ISSUE #776: contextual-decorator: fails for some cases
~**Case 1:** (It doesn't report failures for pipes)~ **Fixed in #754**.

```
@Pipe({
  name: 'app-test'
})
export class TestPipe {
  @Input() someVar: string; // should report a failure, currently it does not.
}
```

**Case 2:** (it doesn't report failures if `ViewChild` decorator is applied to a setter accessor)
```
@Injectable({ // doesn't work for @Pipe also
  providedIn: 'root'
})
export class TestService {
  @ViewChild(Pane)  // should report a failure, currently it does not.
  set pane(v: Pane) {
    console.log('v', v);
  }
}
```

<hr>

Also, the `@Attribute` decorator is not handled at the moment. Is it intentional? @mgechev @wKoza

rafaelss95:
This:

 - Fixes a bug introduced in the new version; (Currently it's not reporting failures if the decorator has arguments due to https://github.com/mgechev/codelyzer/blob/c07ac5e085365562301672cef4e59f159b7a84d0/src/contextualDecoratorRule.ts#L66
 - Reports failures for `accessors` with decorators (ex. `@Input`, `@ViewChild`);
 - Reports failures for `parameter properties` decorators (ex. `@Attribute`, `@Host`, etc...);

Fixes #776.

### DIFF
--- a/test/contextualDecoratorRule.spec.ts
+++ b/test/contextualDecoratorRule.spec.ts
@@ -1,6 +1,6 @@
 import { getFailureMessage, Rule } from '../src/contextualDecoratorRule';
-import { AngularClassDecorators, AngularInnerClassDecorators } from '../src/util/utils';
-import { assertAnnotated, assertSuccess } from './testHelper';
+import { AngularClassDecorators } from '../src/util/utils';
+import { assertAnnotated, assertMultipleAnnotated, assertSuccess } from './testHelper';
 
 const {
   metadata: { ruleName }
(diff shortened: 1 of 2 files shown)


rafaelss95:
I've added the `blocked` label until the #797 gets in.

test/contextualDecoratorRule.spec.ts
```diff
@@ -1,502 +1,1065 @@
...
-import { assertAnnotated, assertSuccess } from './testHelper';
+import { AngularClassDecorators } from '../src/util/utils';
+import { assertAnnotated, assertMultipleAnnotated, assertSuccess } from './testHelper';
 
 const {
   metadata: { ruleName }
 } = Rule;
 
-describe(ruleName, () => {
+describe.only(ruleName, () => {
```
mgechev: Would you drop the `.only` part

rafaelss95: Done.

Issue #558, opened 2018-04-09
add new rule to enforce the use of providedIn flag


wKoza:
This is recognized as best practice for providers with Angular 6 (we can now perform tree shaking and remove services that aren't used by your app

rafaelss95:
Related to #135. Btw, any news on integrate material2#validateDecoratorsRule here?

mgechev:
Following the feature request process that [lodash has](https://github.com/lodash/lodash/blob/master/.github/CONTRIBUTING.md#feature-requests), I'm closing the issue and adding `votes needed` label.

Later we can prioritize the feature requests by popularity and include them in a future release.

mohammedzamakhan:
You can assign it to me

mgechev:
GitHub does not allow this but since you're taking it I'll drop `votes needed`.

rafaelss95:
@mgechev @wKoza 
Currently we have the following rules to (dis)allow decorator properties values:

| List of rules |
|:-:|
| `no-host-metadata-property`                                 |
| `no-inputs-metadata-property`                              |
| `no-outputs-metadata-property`                            |
| `no-queries-metadata-property`                            |
| `no-pipe-impure`                                                      |
| `prefer-on-push-component-change-detection`  |
| `use-component-selector`                                       |
| `use-component-view-encapsulation`                    |

... and this list is growing...

What's your opinion about implement something like [**`material2#validateDecoratorsRule`**](https://github.com/angular/material2/blob/master/tools/tslint-rules/validateDecoratorsRule.ts) (it was already mentioned in this and other issues like https://github.com/mgechev/codelyzer/issues/135)?

A default config would be:

```JSON
{
  "rules": {
    "validate-decorators": [
      true, {
        "Component": {
          "changeDetection": "\\.OnPush$",
          "!encapsulation": "\\.None$",
          "!host": ".*",
          "!inputs": ".*",
          "!outputs": ".*",
          "!queries": ".*",
          "selector": ".*"
        },
        "Directive": {
          "selector": ".*"
        },
        "Pipe": {
          "!pure": ".*"
        }
      }
    ]
  }
}
```

... and of course we could allow users to customize it according to their project standards, and why not also their decorators.

mgechev:
Sounds good. I can talk to @jelbourn to converge both rules.

jscharett:
Does this rule allow for a non-singleton services?  I haven't seen an example where provideIn is used for a component, only root/Modules.

PR #814, opened 2019-04-23
feat(rule): add use-injectable-provided-in

mohammedzamakhan:
https://github.com/mgechev/codelyzer/issues/558


test/useProvidedInRule.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+import { assertAnnotated, assertSuccess } from './testHelper';
+
+const {
+  metadata: { ruleName },
+  FAILURE_STRING
+} = Rule;
+
+describe(ruleName, () => {
+  describe('failures', () => {
+    it('should fail if providedIn property is not passed', () => {
```
rafaelss95:
```suggestion
    it('should fail if providedIn property is not set', () => {
```

test/useProvidedInRule.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+      const source = `
+        @Injectable({
+          providedIn: 'root'
+        })
+        class Test {}
+      `;
+      assertSuccess(ruleName, source);
+    });
+
+    it('should succeed if providedIn property is set to a property', () => {
```
rafaelss95:
```suggestion
    it('should succeed if providedIn property is set to a module', () => {
```

rafaelss95:
@mohammedzamakhan I've left some minor comments, please have a look.

rafaelss95:
@mohammedzamakhan Can you just change the commit message to `feat(rule): add use-injectable-provided-in`?

mohammedzamakhan:
@rafaelss95 done 👍

mgechev:
I'll restart travis and merge when the build passes.

rafaelss95:
I investigated the reason behind the build failure and it seems that the problem is with `node-sass` dependency: our _travis_ config (as you may already know) currently runs with Node 8.11.1 and Node 12 (`latest stable`), however, [node-sass is not compatible with version 12 node](https://github.com/sass/node-sass/issues/2632).

The fix should have come in https://github.com/sass/node-sass/commit/23c86596a9bbcf1e72c6871e3a7cf084d7b68496, [but apparently it hasn't yet been published in npm](https://www.npmjs.com/package/node-sass).

PR #866, opened 2019-08-01
docs: update docs with all new 5.0 names

### LINKED ISSUE #865: Documentation not updated with 5.0 names
**Describe the bug**

**Context and configuration**

The documentation http://codelyzer.com/rules/templates-no-negated-async/ says that the rule is called `templates-no-negated-async` but the rule appears to have been renamed to `template-no-negated-async` both in code and the release notes.

**Environment**

 - Version 5.0.1

aboyton:
I'm not sure if this is done correctly as [the contribution guide](https://github.com/mgechev/codelyzer/blob/master/CONTRIBUTING.md#documentation-related-prs) says not to edit the `docs` directory directly, but this seemed to be the only place referencing the old names (other than some that are also in `docs/src/worker.ts`).

Fixes #865


mgechev:
Thanks for the catch!

PR #887, opened 2019-09-04
Add missing recommended rules

YogliB:
These rules are defined in the latest Angular template ([here](https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/workspace/files/tslint.json.template)) and were missing from the recommended rules list.


mgechev:
Thanks for the catch!

Issue #888, opened 2019-09-15
Pseudo events not supported with accessibility rule


Edouardbozon:
**Describe the bug**

Hi, I would like to use accessibility rules [as described here](https://web.dev/accessible-angular-with-codelyzer/) but they don't support pseudo events.

![image](https://user-images.githubusercontent.com/8522558/64920317-75b00380-d7b6-11e9-9009-f354e7ed3d2a.png) 

**Context and configuration**

Please share:
 - Which rule is causing the problem: **template-click-events-have-key-events**

**Expected behavior**

Support of pseudo events for accessibility rules.

**Code**

```ts
<a *ngFor="let name of groupNames"
    [class.selected]="name == selectedGroup.name"
    class="button mat-button filter-button"
    (click)="selectGroup(name)"
    (keyup.enter)="selectGroup(name)">{{name}}</a>
```

lazarljubenovic:
Since we're talking about accessibility, that thing's supposed to be a button, not an anchor.

Edouardbozon:
@lazarljubenovic In this case `selectGroup(name)` do a redirection using the `PlatformLocation` API. IMO the `<a>` element is appropriate.

lazarljubenovic:
It's an anchor tag without `href`, which means it's a "placeholder for a link". Things like "Open in new tab" won't work. You should pre-compute the URL and then use `routerLink`.

Edouardbozon:
@lazarljubenovic Ok noted, thanks for the advice. 

However it doesn't resolve the main problem with pseudo events. Here is an other example without link element : 

```ts
<div class="card-front" (click)="flipCard(person)" (keyup.enter)="flipCard(person)">
  <h3>{{person.name}}</h3>
</div>
```

The rule should cover this case as described [here](https://medium.com/ngconf/new-accessibility-rules-in-codelyzer-v5-0-0-85eec1d3e9bb).

>  A non-interactive element that has a click event, should be accompanied by one of the key events (keyup, keydown or keypress)

But it fails when using pseudo events.

lazarljubenovic:
Yeah, sorry for side-tracking the topic :pray: You're completely correct that it's a bug (or, well, a missing feature) in Codelyzer.

mgechev:
@mohammedzamakhan do you think you can alter the rule `template-click-events-have-key-events` so we can handle this case?

mohammedzamakhan:
yes we can update it to handle it, I will take a look into it

[...]

PR #893, opened 2019-09-18
fix(rule): template-click-events-have-key-events - support pseudo events

mohammedzamakhan:


mohammedzamakhan:
fixes #888

mgechev:
Just published the change in 5.1.1.

PR #961, opened 2020-03-12
fix: add support for tslint 6.0+

alan-agius4:
Tslint 6.0+ adds support for TypeScript 3.8 syntax. 

https://github.com/palantir/tslint/pull/4915


PowerKiKi:
Now that Angular 9.1 shipped with tslint 6 being the new default version, would it be possible to also get a new release of codelyzer in order to avoid the following error on a brand new Angular project ?

```
warning " > codelyzer@5.2.1" has incorrect peer dependency "tslint@^5.0.0".
```

PR #974, opened 2020-08-04
Change relative link to external site to absolute

Yrkki:
Relative links to external sites only lead you to a 404 🤷‍♂️

Switch to absolute maybe?


mgechev:
Nice catch!

